PR #14, opened 2026-03-19
Feature/improve UI

RivkyKlapholtz:


RivkyKlapholtz (commented):
Thanks for the improvements here - especially the Monaco path normalization and the search endpoint.

I reviewed this PR and found a few issues worth addressing before merge:

1) High: Monaco hydration effect currently depends on openFiles, which changes on almost every edit. This re-triggers broad model hydration and repeated file fetches, and can cause TS worker churn plus intermittent import diagnostics. Consider removing openFiles from the effect dependencies and running hydration only on sessionId/fileTree changes with bounded concurrency.

2) Medium: /api/files/{session_id}/search request model does not enforce bounds for max_results / max_hits_per_file. This can be abused (or accidentally set very high) and cause heavy CPU/memory usage. Recommend adding pydantic constraints (Field(ge=1, le=...)) and query length limits.

3) Medium: In search.py, overflow-file scanning can exceed global max_results because full per-file hit lists are appended before enforcing remaining budget. Consider passing a 
emaining cap into overflow scanning and truncating hits before append.

4) Low: There is an unexpected empty file 	askkill in the PR diff. This looks accidental and should be removed.

Happy to re-review after these are addressed.

backend/app/api/files.py
```diff
@@ -17,6 +18,13 @@ class WriteFileRequest(BaseModel):
     content: str
 
 
+class SearchRequest(BaseModel):
+    query: str
+    case_sensitive: bool = False
+    max_results: int = 2000
```
RivkyKlapholtz: Could we add request validation bounds here? max_results/max_hits_per_file currently accept any integer, which can cause very heavy searches under accidental or malicious input. Suggest Field(ge=1, le=...) plus a query-length cap.

backend/app/sandbox/search.py
```diff
@@ -0,0 +1,314 @@
...
+            if total >= max_results:
+                break
+            hits = await _search_overflow_file(
+                rel,
+                abs_path,
+                needle,
+                case_sensitive=case_sensitive,
+                max_hits=max_hits_per_file,
+            )
+            if hits:
```
RivkyKlapholtz, line 274: Potential correctness issue: in the overflow path we append all hits for a file and only then update/check the global 	otal. This can overshoot max_results in one step. Consider passing 
emaining = max_results - total into _search_overflow_file and truncating before append.

frontend/src/components/editor/EditorPanel.tsx
```diff
@@ -101,43 +282,208 @@ export function EditorPanel() {
...
+  
+      await Promise.allSettled(tasks);
     };
+  
+    void hydrateModels();
+  
+    return () => {
+      cancelled = true;
+    };
+  }, [sessionId, fileTree, flattenFiles, openFiles]);
```
RivkyKlapholtz: High-impact perf note: this effect depends on openFiles, so it re-runs while typing and can trigger repeated hydration/fetch work for many files. Consider limiting dependencies to sessionId + ileTree (and guarding with an in-flight map) to avoid TS worker churn and transient diagnostics.

taskkill
RivkyKlapholtz, line 1: This empty file looks accidental (	askkill). Could we remove it from the PR to keep the diff clean?

88roy88: yes please remove it

backend/app/ai/prompts/architecture.py
```diff
@@ -28,5 +29,7 @@
 - Keep it simple — prefer useState/useReducer over external state libraries unless complexity warrants it
 - All file paths are relative to the project root (e.g. src/App.tsx)
 - Include only the files that need to be created or modified
+- Include a barrel re-export file `src/types.ts` that re-exports everything from `src/types/index.ts`
+  so imports like `import { ... } from './types'` resolve reliably in TS/Monaco.
```
88roy88: The AI doesnt care about TS/Monaco :)

backend/app/sandbox/filesystem.py
```diff
@@ -113,7 +113,10 @@ def _build_tree(dir_path: str) -> list[FileEntry]:
             if name.startswith(".") or name in SKIP_DIRS:
                 continue
             abs_path = os.path.join(dir_path, name)
+            # Monaco/TypeScript resolver עובדים עם URI בצורה עקבית (עם `/`).
```
88roy88: please no comments in hebrew :)

88roy88 (commented):
you can close this one, this change will get in in this pr: https://github.com/88roy88/FlowBolt/pull/29

PR #20, opened 2026-03-22
Feature/auth package api

RivkyKlapholtz:


backend/app/ai/agents/build.py
```diff
@@ -45,11 +45,19 @@ def __init__(self, **kwargs) -> None:
...
     # -- Entry point --
 
     @observe(name="build-agent-run")
-    async def run(self, content: str, case_ids: list[str] | None = None) -> None:
+    async def run(
+        self,
+        content: str,
+        case_ids: list[str] | None = None,
+        *,
+        package_api_authorization: str | None = None,
```
88roy88: why are we taking `self._package_api_authorization` both in the init and here? lets keep it only in the init?

backend/src/flow44/api/chat.py
```diff
@@ -83,17 +83,24 @@ async def _receive_actions() -> None:
                 selected_model: str | None = data.get("model")
                 case_ids: list[int] = data.get("caseIds") or []
 
+                raw_pkg_auth = data.get("packageApiAuthorization")
```
88roy88: I think we should do this in the handshake and not in each action.

backend/src/flow44/api/chat.py
```diff
@@ -83,17 +83,24 @@ async def _receive_actions() -> None:
...
                 # Emit user_message event (for frontend history reconstruction)
                 user_event: dict = {"type": "user_message", "content": user_content}
                 if case_ids:
                     from app.api.package_api import _package_search
+
                     case_names: list[str] = []
                     for cid in case_ids:
                         try:
-                            results = await _package_search(str(cid))
+                            results = await _package_search(str(cid), authorization=package_api_authorization)
```
88roy88: I know its me and not you, but I think this step should be part of the Build Agent, can you move it inside?

backend/src/flow44/api/package_api.py
```diff
@@ -18,56 +19,91 @@
 
 router = APIRouter(prefix="/api/package", tags=["package"])
 
+# APIKeyHeader (not Header(alias=...)) so Swagger UI actually sends Authorization on Try it out.
+_package_authorization = APIKeyHeader(name="Authorization", auto_error=False)
```
88roy88: I think we will want the auth to be on all the endpoints. don't we? this way we could also make the projects private to a specific user and allow login to our system only if you are connected to flow. I want this to be an Auth middleware.
read:
`https://fastapi.tiangolo.com/tutorial/middleware`
`https://fastapi.tiangolo.com/advanced/security/http-basic-auth/?h=auth`

backend/src/flow44/api/package_api.py
```diff
@@ -18,56 +19,91 @@
...
 
     try:
-        return await _client().search(query_or_id)
+        return await _client(authorization=authorization).search(query_or_id)
     except PackageApiUpstreamError as e:
-        raise HTTPException(status_code=502, detail=str(e)) from e
+        raise _map_upstream_error(e) from e
+
+
+def _normalize_package_auth(raw: str | None) -> str | None:
```
88roy88: not sure its needed.

backend/src/flow44/api/package_api.py
```diff
@@ -18,56 +19,91 @@
...
 
-def _client() -> PackageApiClient:
-    return PackageApiClient(base_url=settings.PACKAGE_API_BASE_URL)
 
+def _client(*, authorization: str | None) -> PackageApiClient:
+    return PackageApiClient(base_url=settings.PACKAGE_API_BASE_URL, authorization=authorization)
 
-async def _package_search(query_or_id: str) -> list[Any]:
+
+def _map_upstream_error(e: PackageApiUpstreamError) -> HTTPException:
```
88roy88: I prefer us to just return 401 as well. no need this function

backend/src/flow44/api/package_api.py
88roy88: also, we will want to add jwt validation :)

88roy88 (changes requested):

backend/src/flow44/api/package_api.py
RivkyKlapholtz: Long term, I agree we’ll want unified auth across the system. However, the current code is handling Authorization forwarding to the Package API, which is a separate concern. A proper full auth solution would need broader design across HTTP, WebSocket, and project ownership.

backend/src/flow44/api/package_api.py
RivkyKlapholtz: This function is small, but it is still useful:

def _normalize_package_auth(raw: str | None) -> str | None:
    if raw is None:
        return None
    stripped = raw.strip()
    return stripped or None

What it does:

None stays None
a whitespace-only value becomes None
" Bearer abc " becomes "Bearer abc"

That is helpful because it keeps the input clean before passing it forward.

Why it is useful:

prevents forwarding an empty or whitespace-only header
keeps the normalization logic in one place
avoids repeating .strip() or None in multiple endpoints
makes the intent clearer

Is it strictly required?
No. The logic could be written inline, for example:

authorization = authorization.strip() or None if authorization else None

But that is a bit less readable and less reusable.

backend/src/flow44/api/package_api.py
RivkyKlapholtz: This function makes an important distinction:

if the upstream returns 401, we return 401
otherwise we return 502, because this is an upstream/proxy failure, not a client auth failure

Returning 401 for every upstream error would be misleading for cases like 500, 404, 503, or timeout.

The helper function is also useful because it keeps this mapping logic in one place instead of duplicating

**what do you say?**

backend/src/flow44/api/package_api.py
RivkyKlapholtz: Yes as a future direction, but not as an immediate replacement for this code.

JWT validation is part of a full authentication system.
The current code does not try to validate JWTs — it only reads the Authorization header and forwards it to the upstream service.

So:

if this backend is only acting as a proxy to the Package API, local JWT validation may not be needed right now
if the system is supposed to trust the JWT and identify users itself, then proper validation is required

Conclusion: the comment is valid as a roadmap item, but not as a strong criticism of this PR by itself, unless the original requirement was to implement real system auth.

**Noa told me that the focused goal of the current branch is to add authentication with FLAPI, and nothing beyond that.**

backend/src/flow44/ai/agents/build.py
```diff
@@ -40,16 +40,19 @@
 # TODO: maybe split this into two agents and get rid of the approval event stuff?
 # TODO: or think about how to add support for it in the Flow framework.
 class BuildAgent(BaseAgent):
-    def __init__(self, **kwargs: Any) -> None:
+    def __init__(
+        self,
+        *,
```
88roy88, line 45: not sure I understand why the put in after the *. why did you choose this? (might be fine, just want to understand)

backend/src/flow44/ai/agents/build.py
```diff
@@ -40,16 +40,19 @@
...
+        package_api_authorization: str | None = None,
+        **kwargs: Any,
+    ) -> None:
         super().__init__(**kwargs)
         self._state = BuildState(session_id=self.session_id, project_id=self.project_id, model=self.model)
         self._observation_id: str | None = None
-        # Pause/resume for plan approval
```
88roy88, line 48: can we keep this comment for now

backend/src/flow44/ai/agents/build.py
```diff
@@ -180,15 +183,10 @@ async def _accept_plan(self) -> None:
 
     async def _fetch_and_analyze_case(self, package_id: str) -> dict[str, Any] | None:
```
88roy88: we renamed case to data_source (thats why I wanted you merge main into your code :) )

backend/src/flow44/ai/agents/build.py
```diff
@@ -212,7 +210,7 @@ async def _fetch_and_analyze_case(self, package_id: str) -> dict[str, Any] | Non
                 analysis = {
```
88roy88: as the exception says, lets fail the build if we failed to fetch the cubes

backend/src/flow44/integrations/data_source_cases.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import annotations
```
88roy88, line 1: we already have the intergrations folder for this, why another one?

backend/src/flow44/ai/agents/build.py
RivkyKlapholtz: The `*` makes `package_api_authorization` a keyword-only argument.
So it must be passed explicitly as:

```python
BuildAgent(package_api_authorization="Bearer ...", ...)
```

and not positionally.

I chose it for clarity and safety rather than because it is functionally required. Since `BuildAgent` is already keyword-oriented (`**kwargs` passed through to `BaseAgent`), this makes the call sites more explicit, self-documenting, and less error-prone during refactors.

That said, if we prefer a simpler signature, removing the `*` would also work.

backend/src/flow44/ai/agents/build.py
```diff
@@ -69,9 +73,20 @@ async def run(self, content: str, data_source_ids: list[str] | None = None) -> N
...
+            failures = [r for r in results if isinstance(r, Exception)]
+            if failures:
+                await self.emit(
+                    {
+                        "type": "error",
+                        "message": "Build aborted: failed to fetch required data source data.",
+                    }
+                )
+                await self.emit({"type": "phase", "phase": "idle"})
+                return
```
88roy88, line 86: why return and not raise?

backend/src/flow44/ai/agents/build.py
```diff
@@ -41,16 +41,20 @@
 # TODO: maybe split this into two agents and get rid of the approval event stuff?
 # TODO: or think about how to add support for it in the Flow framework.
 class BuildAgent(BaseAgent):
-    def __init__(self, **kwargs: Any) -> None:
+    def __init__(
+        self,
+        *,
+        flapi_api_authorization: str | None = None,
```
88roy88: im wondering if we should call it flapi here or data_source. we want to keep it decoupled. wdyt

backend/src/flow44/ai/agents/build.py
```diff
@@ -69,9 +73,20 @@ async def run(self, content: str, data_source_ids: list[str] | None = None) -> N
         if self._state.data_source_ids:
             await self.emit({"type": "phase", "phase": "fetching_data_sources"})
             results = await asyncio.gather(
-                *[self._fetch_and_analyze_data_source(sid) for sid in self._state.data_source_ids]
+                *[self._fetch_and_analyze_data_source(sid) for sid in self._state.data_source_ids],
+                return_exceptions=True,
```
88roy88: by adding return_exceptions, we just wait for others to complete, I think that we want to fail fast in this case. 
so we also dont need `failures`

backend/src/flow44/ai/agents/build.py
```diff
@@ -223,13 +228,18 @@ async def _fetch_and_analyze_data_source(self, data_source_id: str) -> dict[str,
                 **analysis,
             }
 
+        except LookupError as e:
+            await self.emit({"type": "data_source_error", "message": f"Data source {data_source_id} not found"})
+            raise RuntimeError(f"Data source fetch failed for {data_source_id}: not found") from e
         except FlapiUpstreamError as e:
```
88roy88: remove the specific exception then

backend/src/flow44/ai/agents/build.py
```diff
@@ -223,13 +228,18 @@ async def _fetch_and_analyze_data_source(self, data_source_id: str) -> dict[str,
...
             await self.emit({"type": "data_source_error", "message": f"Failed to fetch data source: {e}"})
-            return None
-        except Exception:
+            raise RuntimeError(f"Data source fetch failed for {data_source_id}: upstream error") from e
+        except RuntimeError:
```
88roy88, line 232: no need to capture

backend/src/flow44/ai/prompts/templates/architecture.jinja2
```diff
@@ -39,6 +39,6 @@ Sample data:
 ```json
 {{ ctx.sample_data_json }}
 ```
-Endpoint: `${API_BASE}/api/data-source/{{ ctx.data_source_id }}/run` (import API_BASE from src/config.ts)
+Endpoint: `${API_BASE}/api/data-source/{{ ctx.data_source_id }}/run` (import API_BASE from src/config.ts) with `Authorization` header from localStorage key `flowbolt.flapiApiToken`
```
88roy88: this is a great question, how will the child app will get the token, not sure that the auth should be at localStorage

88roy88: we might want to inject something.

backend/src/flow44/ai/state.py
```diff
@@ -25,3 +25,7 @@ class BuildState(BaseModel):
     phase: str = "idle"
 
     model_config = {"arbitrary_types_allowed": True}
+
+
+# Pydantic v2: ensure forward refs are resolved for runtime model creation.
+BuildState.model_rebuild()
```
88roy88: I dont think this is needed.

backend/src/flow44/api/chat.py
```diff
@@ -73,13 +83,17 @@ async def _forward_events() -> None:
...
+    async def _receive_actions() -> None:  # noqa: C901, PLR0912, PLR0915
+        flapi_api_authorization: str | None = None
         while True:
             raw = await websocket.receive_text()
             data = json.loads(raw)
             msg_type = data.get("type")
 
-            if msg_type == "message":
+            if msg_type == "auth":
+                raw_auth = data.get("flapiApiAuthorization")
```
88roy88: same question about the naming

backend/src/flow44/api/flapi_api.py
88roy88: lets remove all references to flapi, including file name, all should be data sources (we will call it flapi in the integration only)

backend/src/flow44/db/database.py
```diff
@@ -45,12 +45,171 @@ async def reset() -> None:
...
 
-async def init_db() -> None:
+async def init_db() -> None:  # noqa: C901
     from sqlmodel import SQLModel  # noqa: PLC0415
 
     import flow44.db.chat  # noqa: F401, PLC0415
     import flow44.db.events  # noqa: F401, PLC0415
     import flow44.db.project  # noqa: F401, PLC0415
 
+    def _ensure_sqlite_projects_columns(conn: Any) -> None:
```
88roy88: what is this? we dont need this migrations anymore

backend/ruff.toml
```diff
@@ -132,3 +132,4 @@ line-ending = "auto"
 "tests/*" = ["S101"]
 "*/test_*.py" = ["S101"]
 "*_test.py" = ["S101"]
+"src/flow44/db/database.py" = ["S608"]
```
88roy88: remove after removing migration code

88roy88 (approved):

PR #30, opened 2026-03-26
feat: postgreSql integration

shalom-raichman:


backend/pyproject.toml
```diff
@@ -5,13 +5,15 @@ description = "AI Web App Builder - Backend"
...
     "aiosqlite>=0.20.0",
+    "asyncpg>=0.31.0",
     "boto3>=1.42.68",
     "fastapi>=0.115.6",
     "greenlet>=3.3.2",
     "httpx>=0.28.0",
     "jinja2>=3.1.0",
     "langfuse==2.59.7",
     "litellm>=1.55.10,<1.82.3",
+    "psycopg2>=2.9.11",
```
88roy88: use psycopg2-binary

backend/src/flow44/db/database.py
```diff
@@ -22,9 +24,10 @@ def get_engine(url: str | None = None) -> AsyncEngine:
     # TODO: this is specific for SQLite, remove when migrating to Postgres
```
88roy88, line 47: when using pg, we need to add connection pool settings

backend/pyproject.toml
```diff
@@ -5,13 +5,15 @@ description = "AI Web App Builder - Backend"
 requires-python = ">=3.12"
```
88roy88, line 5: in the example.env and config.py change default values to pg

backend/pyproject.toml
```diff
@@ -5,13 +5,15 @@ description = "AI Web App Builder - Backend"
...
 dependencies = [
```
88roy88, line 6: in docker compose add pg as dependecy.

backend/pyproject.toml
```diff
@@ -5,13 +5,15 @@ description = "AI Web App Builder - Backend"
...
     "aiosqlite>=0.20.0",
```
88roy88, line 7: and we are missing alembic

88roy88 (changes requested):

backend/src/flow44/db/database.py
```diff
@@ -11,6 +11,8 @@ def _get_async_url() -> str:
     url = flow44.config.settings.DATABASE_URL
```
88roy88: we should probably remove `await init_db()` and use alembic somewhere

backend/migrations/versions/7c1991be16e2_initial_migration.py
```diff
@@ -0,0 +1,32 @@
+"""Initial migration
```
88roy88: Personally, I like it that the migration start with date so it's always ordered, it's a config in the alembic.ini

backend/migrations/versions/7c1991be16e2_initial_migration.py
```diff
@@ -0,0 +1,32 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "7c1991be16e2"
+down_revision: Union[str, Sequence[str], None] = None
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
```
88roy88: This is not good :) it means that it didn't read the sqlmodels, this file should be filled with create table commands

backend/src/flow44/config.py
```diff
@@ -20,13 +20,17 @@ class Settings(BaseSettings):
     NSJAIL_BIN: str = "/usr/bin/nsjail"
     AI_MODEL: str = "bedrock/us.anthropic.claude-sonnet-4-6"
     OLLAMA_BASE_URL: str = "http://localhost:11434"
-    DATABASE_URL: str = "sqlite:///./ai_builder.db"
+    DATABASE_URL: str = ""
```
88roy88: Let's not give default value, the user must specifiy it

backend/src/flow44/config.py
```diff
@@ -47,7 +51,11 @@ class Settings(BaseSettings):
     LANGFUSE_SECRET_KEY: str = ""
     LANGFUSE_HOST: str = "https://cloud.langfuse.com"
 
-    model_config = SettingsConfigDict(env_prefix="AIB_", env_file=".env", extra="ignore")
+    model_config = SettingsConfigDict(
+        env_prefix="AIB_",
+        env_file=[".env", str(_BACKEND_ROOT.parent / ".env")],
```
88roy88: Why. Let's keep the .env local to the backend

backend/src/flow44/db/database.py
```diff
@@ -46,11 +61,11 @@ async def reset() -> None:
...
-
+    """Initialize database connection. Metadata creation is now handled by Alembic."""
     import flow44.db.chat  # noqa: F401, PLC0415
     import flow44.db.events  # noqa: F401, PLC0415
     import flow44.db.project  # noqa: F401, PLC0415
 
+    # We just ping the engine to ensure connectivity
     async with get_engine().begin() as conn:
-        await conn.run_sync(SQLModel.metadata.create_all)
+        await conn.execute(text("SELECT 1"))
```
88roy88, line 78: Remove

88roy88 (changes requested):

backend/src/flow44/config.py
shalom-raichman: The default value is an empty string

backend/src/flow44/db/database.py
shalom-raichman: don't we want to ensure connectivity ?

backend/src/flow44/alembic/versions/2026_03_29_1207-9d6eeadce019_initial_migration.py
```diff
@@ -0,0 +1,66 @@
+"""Initial migration
```
88roy88, line 1: The filename is still wrong ;)

backend/src/flow44/config.py
88roy88: No default value at all! (Remove the ="")

backend/src/flow44/db/database.py
88roy88: Even if so, not like this.

88roy88 (changes requested):

backend/test.env
```diff
@@ -0,0 +1,2 @@
+# Test Environment Configuration
+AIB_DATABASE_URL="sqlite+aiosqlite:///:memory:"
```
88roy88: great! so we can probably remove some lines from `setup_test_db`

backend/tests/conftest.py
88roy88: we dont need this line anymore if we have the test.env

backend/src/flow44/config.py
```diff
@@ -41,13 +45,18 @@ class Settings(BaseSettings):
...
 
     # Langfuse (optional — set public/secret key to enable)
     LANGFUSE_PUBLIC_KEY: str = ""
     LANGFUSE_SECRET_KEY: str = ""
     LANGFUSE_HOST: str = "https://cloud.langfuse.com"
 
-    model_config = SettingsConfigDict(env_prefix="AIB_", env_file=".env", extra="ignore")
+    model_config = SettingsConfigDict(
+        env_prefix="AIB_",
+        env_file=(".env", "test.env"),
```
88roy88: no, never in prod load test.env. also, move test.env to under tests/

backend/src/flow44/db/database.py
```diff
@@ -46,11 +61,11 @@ async def reset() -> None:
 
 
 async def init_db() -> None:
-    from sqlmodel import SQLModel  # noqa: PLC0415
-
+    """Initialize database connection. Metadata creation is now handled by Alembic."""
     import flow44.db.chat  # noqa: F401, PLC0415
```
88roy88: do we need this imports here now?

88roy88 (changes requested):

backend/test.env
shalom-raichman: actually we can't because in the setup_test_db we use dynamic url so the test can run in parallel

backend/tests/conftest.py
shalom-raichman: These lines enable running the tests in parallel

backend/src/flow44/alembic/env.py
```diff
@@ -0,0 +1,84 @@
+import asyncio
```
88roy88, line 1: move the migration folder to be under src and rename it to alembic.

backend/src/flow44/db/database.py
```diff
@@ -11,20 +11,35 @@ def _get_async_url() -> str:
     url = flow44.config.settings.DATABASE_URL
```
88roy88: this function should take 
DB_NAME
DB_HOST
DB_USERNAME ...
and build the url

PR #60, opened 2026-04-13
feat: add debounce to data source search

### LINKED ISSUE #53: debounce for data-sources search

OriCohen05:
Closes #53

## Summary
- Debounces the `searchDataSources` API call by 300ms so requests are only fired after the user stops typing
- `isLoading` is set immediately on each keystroke so the UI stays responsive while the debounce is pending
- Pending timers are cleared on each new keystroke to prevent stale requests

## Test plan
- [x] Type in the data source search field and verify only one API request fires after a 300ms pause
- [x] Verify clearing the input hides the dropdown immediately without triggering a request
- [x] Verify selecting a result still works correctly

### DIFF
--- a/frontend/src/components/chat/DataSourceSelector.tsx
+++ b/frontend/src/components/chat/DataSourceSelector.tsx
@@ -21,7 +21,6 @@ export function DataSourceSelector({ isOpen }: DataSourceSelectorProps) {
   const removeDataSource = useChatStore((s) => s.removeDataSource);
   const inputRef = useRef<HTMLInputElement>(null);
   const dropdownRef = useRef<HTMLDivElement>(null);
-  const debounceRef = useRef<ReturnType<typeof setTimeout> | null>(null);
 
   useEffect(() => {
     const handleClickOutside = (event: MouseEvent) => {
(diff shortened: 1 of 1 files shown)


frontend/src/components/chat/DataSourceSelector.tsx
```diff
@@ -38,24 +39,27 @@ export function DataSourceSelector({ isOpen }: DataSourceSelectorProps) {
...
-    try {
-      const sources = await searchDataSources(searchQuery);
-      setResults(sources.slice(0, 10));
-    } catch (err) {
-      console.error('Failed to search data sources:', err);
-      setResults([]);
-    } finally {
-      setIsLoading(false);
-    }
+    debounceRef.current = setTimeout(async () => {
```
88roy88: A better approach is to debounce the query state with useEffect (or a small useDebounce hook).
also There’s no cleanup for the timeout on unmount / There’s a potential race condition where older requests can overwrite newer results

88roy88 (changes requested):

frontend/src/components/chat/DataSourceSelector.tsx
OriCohen05: Fixed it. I wonder when to fire the loading icon since now its loading on request before the delay

frontend/src/components/chat/DataSourceSelector.tsx
88roy88, line 114: here `Searching...` and in `No data sources found` can you add i18n support for Hebrew?

OriCohen05: Done

PR #62, opened 2026-04-13
add error logging to follow-up agent, react loop, and litellm provider

shalom-raichman:


backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -184,14 +184,36 @@ async def run(self, content: str) -> None:
...
-        answer = await react_flow.react_loop(
-            messages=messages,
-            system_prompt=system_prompt,
-            tools=self._executor,
-            model=self.model,
-            metadata_fn=lambda step: self._llm_metadata(f"followup-{step}"),
-            emit_fn=self._emit_react_step,
-        )
+
+        try:
```
88roy88: not sure we need this here if we have it in both the provider and in the react flow.


btw, I guess you also added it manually in the code in NH. did we get the errors from 80 and we know what they did wrong ? :)

backend/src/flow44/ai/core/provider.py
```diff
@@ -58,16 +61,28 @@ async def complete_chat_with_tools(
...
-        model=resolved_model,
-        messages=full_messages,
-        api_base=settings.AI_BASE_URL,
-        api_key=settings.AI_API_KEY,
-        tools=tools,
-        tool_choice=tool_choice,
-        stream=False,
-        metadata=metadata or {},
-    )
+    try:
```
88roy88: I would make a decorator or something out of it and put it on all 3 functions in this file. btw.

backend/src/flow44/ai/agents/followup/agent.py
shalom-raichman: I didn't added it manually :(

88roy88 (approved):

PR #64, opened 2026-04-14
feat: open home page on app entry instead of recent project

### LINKED ISSUE #55: defaults to home page

OriCohen05:
Closes #55

### DIFF
--- a/frontend/src/components/layout/AppShell.tsx
+++ b/frontend/src/components/layout/AppShell.tsx
@@ -164,7 +164,7 @@ export function AppShell() {
       const hasMessages = messages.length > 0;
       setProjectHasMessages(currentProject.id, hasMessages);
     }
-  }, [historyLoaded, messages.length, currentProject?.id]);
+  }, [historyLoaded, messages.length, currentProject?.id, currentProject]);
 
   const handleBottomResize = useCallback((delta: number) => {
     setBottomHeight((h) =>
--- a/frontend/src/components/layout/AppShell.tsx
+++ b/frontend/src/components/layout/AppShell.tsx
@@ -50,17 +50,6 @@ function saveSidebarPinned(pinned: boolean) {
   } catch {}
 }
 
-function getProjectHasMessages(projectId: string): boolean | null {
-  try {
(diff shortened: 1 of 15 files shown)


frontend/src/components/layout/AppShell.tsx
```diff
@@ -90,16 +116,46 @@ export function AppShell() {
...
+  );
+
+  // Create an "Untitled" project, navigate to it, then send the first message.
+  // Called when the user submits a prompt from the home screen (no active project).
+  const handleCreateAndSend = useCallback(
+    async (message: string) => {
+      await createProject(t("sidebar.untitledProject"));
+      const session = useSessionStore.getState();
+      if (session.currentProject) {
+        window.location.hash = `#/project/${session.currentProject.id}`;
```
OriCohen05: q: should we use react useNavigate instead? I feel like it's a lazy solution but it also work

frontend/src/components/layout/AppShell.tsx
```diff
@@ -277,17 +404,28 @@ export function AppShell() {
...
+                    !currentProject ? handleCreateAndSend : undefined
+                  }
+                />
               </div>
               <div className="flex flex-wrap gap-2 justify-center">
-                {['A dashboard with charts', 'A todo app with drag & drop', 'A landing page with animations'].map((hint) => (
+                {[
+                  t("onboarding.hintDashboard"),
+                  t("onboarding.hintTodoApp"),
+                  t("onboarding.hintLandingPage"),
```
OriCohen05, line 431: If you prefer not to legitimate hebrew prompts I totally understand, I just hate magic strings

frontend/src/components/layout/AppShell.tsx
88roy88: great q, afaik we don't use React router currently. im down for it though if we have a use case for it - currently we are kinda single page app.

frontend/src/components/layout/AppShell.tsx
88roy88: we should support hebrew

PR #67, opened 2026-04-15
Rivky/48/monaco more options

RivkyKlapholtz:


backend/src/flow44/api/files.py
```diff
@@ -70,6 +80,58 @@ async def put_file_content(sandbox: Annotated[PnpmSandbox, SandboxDep], body: Wr
         raise HTTPException(status_code=403, detail=str(exc)) from None
 
 
+@router.post("/entry")
```
88roy88: not sure im a fan of `entry` maybe `project_file`? or something like that?

RivkyKlapholtz: entry is technically fine, but I agree it feels a bit vague.

I would not use project_file if the API also handles folders, rename, delete, upload, because then the name becomes misleading.

Which of the following is better?:
path
filesystem-entry
file-node
project-item

backend/src/flow44/sandbox/filesystem_mixin.py
```diff
@@ -27,6 +28,32 @@ async def write_file(self, path: str, content: str) -> None:
...
+
+    async def create_file(self, path: str, content: str = "") -> None:
+        full = self._safe_path(path)
+        if os.path.exists(full):  # noqa: ASYNC240
+            raise FileExistsError(path)
+        os.makedirs(os.path.dirname(full), exist_ok=True)
+        with open(full, "x", encoding="utf-8") as f:  # noqa: ASYNC230
+            f.write(content)
+
+    async def rename_path(self, source_path: str, destination_path: str) -> None:
```
88roy88: this should be rename file.

backend/src/flow44/sandbox/filesystem_mixin.py
```diff
@@ -27,6 +28,32 @@ async def write_file(self, path: str, content: str) -> None:
...
 
+    async def write_binary_file(self, path: str, content: bytes) -> None:
+        full = self._safe_path(path)
+        if os.path.exists(full):  # noqa: ASYNC240
+            raise FileExistsError(path)
+        os.makedirs(os.path.dirname(full), exist_ok=True)
+        with open(full, "xb") as f:  # noqa: ASYNC230
+            f.write(content)
+
+    async def create_file(self, path: str, content: str = "") -> None:
```
88roy88, line 39: why we work with str and not Path in all those functions btw?

88roy88 (changes requested):

backend/src/flow44/api/files.py
88roy88: it dosnt handle folders.

backend/src/flow44/api/files.py
88roy88: its kinda weird we have both, /content and /project_file
do we have a tag for all of this (i.e /file/??? )

frontend/src/locales/he.json
```diff
@@ -74,7 +74,20 @@
...
     "noMatchingFiles": "אין קבצים תואמים",
-    "searchShortcutTitle": "חיפוש בקבצים (Ctrl+Shift+F)"
+    "searchShortcutTitle": "חיפוש בקבצים (Ctrl+Shift+F)",
+    "createFile": "צור קובץ",
+    "uploadFiles": "העלה קבצים",
+    "renameFile": "שנה שם",
+    "deleteFile": "מחק",
+    "deletedTab": "קובץ נמחק (הלשונית נשארת פתוחה)",
+    "readOnlyUntilFirstAiResponse": "עריכה תופעל רק אחרי שהבנייה הראשונית של ה-AI הושלמה.",
+    "readOnlyWhileAiWorking": "העריכה נעולה זמנית בזמן שה-AI עובד.",
```
88roy88: צריך לוודא שכל העברית אנגלית הזה נראה סבבה בפרונט ושהימין-לשמאל נראה סבבה

88roy88 (changes requested):

frontend/src/components/editor/FileTree.tsx
```diff
@@ -1,7 +1,11 @@
 import { useFilesStore } from '../../stores/files';
```
88roy88, line 1: this file became huge! +500 new lines. can we split?

88roy88 (changes requested):

88roy88 (approved):

PR #70, opened 2026-04-16
feat: created pre push hook with some commands, we might want to auto…

OriCohen05:
…mate it a little bit more


88roy88:
im against pre hooks :)

PR #73, opened 2026-04-19
feat: integrate white changes into black

shalom-raichman:
- added some envs and args that isn't exists in the white yet but hopefully won't brake anything
- limit the range of the ports killing !!
- add sqlite option in the example.env


backend/.npmrc
```diff
@@ -0,0 +1,3 @@
+registry=${NPM_REGISTRY}
```
88roy88: I dont think we need this file here as well. (verify me)

backend/pnpm-project-template/.npmrc
```diff
@@ -0,0 +1,3 @@
+registry=${NPM_REGISTRY}
+strict-ssl=${NPM_STRICT_SSL}
+audit=${NPM_AUDIT}
```
88roy88, line 3: do we really need the `audit` flag?

backend/App.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ARG UV_INDEX_NAME=default
+ARG UV_INDEX_DEFAULT=true
+ARG UV_INSECURE_HOST1
+ARG UV_INSECURE_HOST2
+
+USER root
+WORKDIR /app
+
+# Copy .npmrc if it exists in the build context
+COPY .npmrc* ./
```
88roy88: not sure why needed for backend

frontend/nginx/nginx.conf
```diff
@@ -0,0 +1,82 @@
+# Map for webSocket connection upgrades
```
88roy88, line 1: lets remove the unused nginx we have then. and adjust the compose

88roy88 (changes requested):

backend/pnpm-project-template/.npmrc
shalom-raichman: no, but it make the install faster

docker-compose.yml
```diff
@@ -15,18 +15,18 @@ x-minio-shared: &minio-shared
   S3_BUCKET_NAME: &s3-bucket flow44
 
 services:
-  frontend:
+  base-frontend.base:
```
88roy88: why the name change?

shalom-raichman: I had some fistuninm with it, apparently we don't need to change the names

docker-compose.yml
```diff
@@ -84,11 +84,12 @@ services:
     restart: unless-stopped
 
   nginx:
```
88roy88: do we need both nginx and frontend service? I thought that the frontend is the nginx now, isnt it?

88roy88 (changes requested):

docker-compose.yml
shalom-raichman: You right, I thought it was separate on purpose, I will fix it right up

docker-compose.yml
```diff
@@ -61,7 +50,9 @@ services:
       start_period: 5s
       retries: 3
     networks:
-      - internal
+      internal:
+        aliases:
+          - base-backend.base
```
88roy88, line 55: can we stop with the base...

shalom-raichman: No, because base-backend.base is how the backend configured in the nginx.conf and if there is no aliases for it the frontend couldn't connect to the backend

PR #74, opened 2026-04-19
Nadav/map mock

Nadav23:


mocks/map-mock/types/visit-map.types.ts
```diff
@@ -0,0 +1,260 @@
+import type { ReactNode } from "react";
```
88roy88, line 1: please add in the flapi mock a package that returns map data (in the right format, like its coming from real flapi). and add a usage example to render the map somewhere (even here in dummy index.html + app.tsx using the flapi data)

88roy88 (changes requested):

mocks/map-mock/types/visit-map.types.ts
Nadav23: I had hard time trying to get the data from the flapi mock so I created it and used it inside the mock

PR #76, opened 2026-04-20
Rivky/fix rg search

RivkyKlapholtz:


backend/Dockerfile
```diff
@@ -31,6 +31,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     libnl-route-3-200 \
     curl \
     ca-certificates \
+    ripgrep \
```
88roy88, line 34: I guess this is the only change, but you created this branch from your other branch so it shows a lot of changes.
👍  lgtm once we merge the other one or you create this branch from main

88roy88 (changes requested):

backend/Dockerfile
RivkyKlapholtz: There are actually more changes, I added tests that would have caught the bug that the search doesn't work...

You will be able to see the changes more clearly once we merge rivky/48/monaco-more-options and pull from master.

PR #79, opened 2026-04-23
Rivky/fix lazy load code on project switch

RivkyKlapholtz:


frontend/src/stores/chat.ts
```diff
@@ -208,14 +208,16 @@ export const useChatStore = create<ChatState>((set, get) => ({
       set({ messages: [], isStreaming: false, historyLoaded: false, buildCompleted: false, ...RESET_STATE });
 
       const socket = getChatSocket(projectId);
-      const handler = createSendMessageHandler(set, get, () => detachHandler(socket, handler));
-      attachHandler(projectId, socket, handler);
+      const replayHandler = createSendMessageHandler(set, get, () => {}, { replay: true });
```
88roy88, line 211: this is a very interesting change, can you explain it?
(I had a debate exactly about this part with Royz last week)

88roy88 (approved):

frontend/src/stores/chatHandlers.ts
```diff
@@ -9,6 +9,7 @@ type SetState = (
 ) => void;
 
 let _skipMessages = false;
```
88roy88, line 11: _skipMessages is global, so not sure the solution with the replay will work with the above change

PR #80, opened 2026-04-23
Rivky/fix rg search clean

RivkyKlapholtz:


backend/pyproject.toml
```diff
@@ -57,16 +57,16 @@ build-backend = "hatchling.build"
...
+ruff-dry = "ruff format --check --config ruff.toml src/ && ruff check --no-fix --config ruff.toml src/"
+ruff-dry-all = "task ruff-dry"
 
-mypy = "mypy --config-file mypy.toml \"$@\""
-mypy-all = "task mypy src/"
+mypy = "mypy --config-file mypy.toml src/"
+mypy-all = "task mypy"
 
-lint = "bash -c 'task mypy \"$@\" && task ruff \"$@\"' --"
-lint-all = "task lint src/"
```
88roy88, line 69: ???

backend/src/flow44/sandbox/search_mixin.py
```diff
@@ -119,5 +102,56 @@ async def grep(  # noqa: C901, PLR0912, PLR0913
...
+            lines: list[str] = []
+            async for line in self.exec(cmd):
+                lines.append(line)
+        except Exception as e:
+            logger.warning("Failed to execute ripgrep command: %s", e, exc_info=True)
+            raise SearchToolError(f"Failed to execute ripgrep command: {e}") from e
+        output = "".join(lines)
+        if not output.strip():
+            logger.warning("Ripgrep returned no output. Command: %s", cmd)
+            raise SearchToolError("Ripgrep returned no output")
```
88roy88: is returning nothing an exception?

backend/pyproject.toml
```diff
@@ -57,16 +57,16 @@ build-backend = "hatchling.build"
...
+lint = "task mypy && task ruff"
+lint-all = "task lint"
```
shalom-raichman, line 69: check if there is way to make it work on windows, and if not ...

backend/pyproject.toml
shalom-raichman: Taskipy appends extra args directly to the command string — it never actually used $@ for arg forwarding

backend/src/flow44/sandbox/search_mixin.py
shalom-raichman: fixed

88roy88 (approved):

PR #84, opened 2026-05-17
fix: white screen when pressing "start over" button

shalom-raichman:


backend/src/flow44/api/chat.py
```diff
@@ -184,7 +184,9 @@ async def _receive_actions() -> None:  # noqa: C901, PLR0912, PLR0915
 
                 elif action == "reject":
                     await delete_pending_plan(project_id)
-                    await emit_event(project_id, {"type": "plan_rejected"})
+                    await emit_event(
+                        project_id, {"type": "plan_rejected", "overview": state.user_overview.model_dump()}
```
shalom-raichman, line 188: Remove start over button completely, clean up the logic

PR #97, opened 2026-05-26
2/4 — Dev watch scoping and gitignore

morbargig:
## Summary
**PR 2/4** — Dev watch scoping and gitignore (merge after #99 ✅).

- Add `.pnpm-store/` to `.gitignore`
- Keep `CLAUDE.md` ignored (unchanged from main)
- Scope Vite watcher to frontend-only paths
- Add `nodemon.json` for flapi-mock watch isolation

**1 commit** · base: `main`

## Merge order
#99 ✅ → **this PR** → #105 → #96 / #106

## Test plan
- [ ] `make dev-backend` / `dev-frontend` / `dev-mocks` in separate terminals
- [ ] `.pnpm-store` not tracked by git
- [ ] `CLAUDE.md` remains gitignored

PR #96, opened 2026-05-26
2/2 — Generated app routing, preview/export, and template guards

morbargig:
## Summary
**PR 2/2** — Generated app routing, preview/export, and template guards.

**4 commits** (reordered for review):

1. Shared preview/export path helpers
2. Vite base paths + full dist publish with SPA fallback
3. Template routing helpers + platform file guard during AI execution
4. Opt-in React Router support, routing-aware AI prompts, merge planner cleanup

Stacks on [#105](https://github.com/88roy88/FlowBolt/pull/105) (`morbargig/94-cancel-generation`), which now includes agent alive polling (`GET /api/iaagent/{id}/alive`), `isAgentAlive` input blocking, and reconnect UI sync.

## Merge order
#99 ✅ → #97 ✅ → #105 → **this PR**

## Test plan
- [ ] Single-page app preview + publish under nested paths
- [ ] Multi-page React Router app — preview, publish, deep route refresh
- [ ] AI cannot edit template platform files or vite.config.ts
- [ ] After refresh mid-generation, chat input stays blocked and live progress matches server (`/iaagent/.../alive`)

PR #105, opened 2026-05-27
1/2 — Fix stale thinking UI after backend reload

morbargig:
## Summary
**PR 1/2** — Frontend-only fix for the chat UI staying in a “thinking” state when generation is no longer running.

- Reset in-progress agent UI when the chat WebSocket closes (e.g. Python backend reload)
- Clear stale state after replaying interrupted runs from history
- Use `isAgentWorking` so progress UI matches active phases (`exploring`, `fixing`, etc.)
- Refactor phase → progress/label mapping to switch-case helpers; `idle` maps to 0% progress (no transient 3% placeholder)

No backend changes. No stop/cancel control.

## Merge order
#99 ✅ → #97 ✅ → **this PR** → #96

## Test plan
- [ ] Start generation → restart backend → UI leaves thinking/spinner state
- [ ] Refresh page mid follow-up run → no stuck exploring/progress UI
- [ ] Normal generation, plan approval, and follow-up flows unchanged
- [ ] Progress bar appears once a real phase arrives (brief gap after send is OK)


frontend/src/services/websocket/chat.ts
```diff
@@ -39,7 +39,11 @@ export function getChatSocket(projectId: string): ChatSocket {
         handlers.forEach((h) => h(msg));
       } catch {}
     },
-    undefined,
+    () => {
```
88roy88: ?

frontend/src/stores/chatHandlers.ts
```diff
@@ -16,6 +16,55 @@ export function setReplayMode(replay: boolean) {
 
 let _lastEventTs = 0;
 
+const TRANSIENT_RESET: Partial<import('./chat').ChatState> = {
```
88roy88: why the inline imports!!!

frontend/src/stores/chatHandlers.ts
```diff
@@ -16,6 +16,55 @@ export function setReplayMode(replay: boolean) {
...
+): boolean {
+  if (isHistoryRunComplete(events)) return false;
+  const state = get();
+  if (state.agentPhase === 'awaiting_approval' && state.planOverview) return false;
+  set(TRANSIENT_RESET);
+  return true;
+}
+
+const ACTIVE_AGENT_PHASES: import('../types').AgentPhase[] = [
+  'classifying',
```
88roy88: is classifying still a real phase? when?

frontend/src/stores/chatHandlers.ts
```diff
@@ -16,6 +16,55 @@ export function setReplayMode(replay: boolean) {
...
+
+/** Clear stale in-progress UI after replaying events from an interrupted run. */
+export function finalizeHistoryReplayState(
+  set: SetState,
+  get: GetState,
+  events: Array<{ type?: string }>,
+): boolean {
+  if (isHistoryRunComplete(events)) return false;
+  const state = get();
+  if (state.agentPhase === 'awaiting_approval' && state.planOverview) return false;
```
88roy88: I dont like the `awaiting_approval` text const everywhere, we can create a const for events where the agent is awaiting and when is not, even better, for the state create a computed field with is_agent_awaits or something

88roy88 (changes requested):

frontend/src/stores/chatHandlers.ts
morbargig: it types import.
when it's build it will disappear   



done

frontend/src/stores/chatHandlers.ts
morbargig: Image

Image


the IDE know how to type check it...
i uses prefer this method because than i can skip that unknown Enum or const that sometimes you have no idea where to find.

but i will add enum.

frontend/src/stores/chatHandlers.ts
morbargig: thanks, I did clear this phase

morbargig:
i fix all cr comments

frontend/src/services/websocket/chat.ts
```diff
@@ -39,7 +46,7 @@ export function getChatSocket(projectId: string): ChatSocket {
         handlers.forEach((h) => h(msg));
       } catch {}
     },
-    undefined,
+    onSocketDisconnect,
```
88roy88: I would just put here `notifyChatConnectionLost` without reassigning

frontend/src/services/websocket/reconnecting.ts
```diff
@@ -3,6 +3,11 @@ export function getWsBase(): string {
   return `${protocol}//${window.location.host}`;
 }
 
+/**
+ * @param onClose Called on every WebSocket `close`, including transient closes
```
88roy88: remove. in general I try to reduce the amount of unnecessary comments

frontend/src/services/websocket/reconnecting.ts
```diff
@@ -3,6 +3,11 @@ export function getWsBase(): string {
   return `${protocol}//${window.location.host}`;
 }
 
+/**
```
88roy88: same.

frontend/src/types/index.ts
```diff
@@ -81,6 +80,9 @@ export type AgentPhase =
   | 'exploring'
   | 'complete';
 
+/** Each phase key must map to itself (`idle` → `'idle'`, not any other phase). */
+export type AgentPhaseConstMap = { readonly [K in AgentPhase]: K };
```
88roy88, line 83: whats the point?

frontend/src/components/layout/GlobalProgress.tsx
```diff
@@ -1,22 +1,25 @@
...
-    agentPhase === 'fetching_data_sources' ? 10 :
-    agentPhase === 'designing' ? 15 :
-    agentPhase === 'planning' ? 30 :
-    agentPhase === 'awaiting_approval' ? 35 :
-    agentPhase === 'executing' ? 40 + (tasks.filter((t) => t.status === 'completed').length / Math.max(tasks.length, 1)) * 55 :
-    agentPhase === 'fixing' ? 50 :
-    agentPhase === 'exploring' ? 50 :
-    agentPhase === 'complete' ? 100 : 0;
+    !working ? 0 :
+    agentPhase === AGENT_PHASE.idle ? 3 :
```
88roy88: why idle is progress?

frontend/src/components/chat/PromptInput.tsx
```diff
@@ -59,20 +60,21 @@ export function PromptInput() {
...
   const placeholder = !projectId
     ? t('chat.placeholder.selectProject')
-    : agentPhase === 'awaiting_approval'
+    : awaitingPlan
       ? t('chat.placeholder.reviewPlan')
       : isBusy
         ? t('chat.placeholder.working')
         : t('chat.placeholder.default');
 
   const busyLabel =
```
88roy88, line 71: can we use case match here?

frontend/src/types/index.ts
```diff
@@ -71,7 +71,6 @@ export interface DataSourceSearchRecord {
 // Agent types
 export type AgentPhase =
```
88roy88, line 72: can we have one source of truth for it (maybe just take it from the store)

88roy88 (changes requested):

88roy88:
for me to go over all comments and resolve them (only I resolve my comments) write done on them if you implemented them

frontend/src/types/index.ts
morbargig: i can remove the comments.

i created like enum but the keys and the values must be the same.

Enum.one = "one" // it will always have the same value to the some keys and the enum will be currect and easy to read with the lowercase 

and not like 

Enum.ONE = "one"

frontend/src/services/websocket/reconnecting.ts
morbargig: i think it's help the ai on dev to understand the project.

i will remove it, lets talk about those use cases later.

morbargig: i really thinks those comments are for the best,
this is not the template it for the base44 app, so in prod you will not see this comments.
but as you are working and the ai working this comments clarify about the exec logic and flow in the WebSocket complexity

frontend/src/types/index.ts
morbargig: it's have one source off true. this one as a type and than object implementation  if there are not sync it will cause compiler errors.

about this setup a 100% recommend to keep as an expert in typescript

frontend/src/services/websocket/chat.ts
morbargig: done

frontend/src/stores/chatHandlers.ts
morbargig: done

frontend/src/stores/chatHandlers.ts
morbargig: done

frontend/src/stores/chatHandlers.ts
morbargig: done

frontend/src/services/websocket/chat.ts
morbargig: done

frontend/src/services/websocket/reconnecting.ts
morbargig: done

frontend/src/components/layout/GlobalProgress.tsx
morbargig: its a an early phase because the classifying is gone to give some indication something is started.

morbargig: i change it back to be 0 %
and i convert it to switch case

morbargig: done

frontend/src/components/chat/PromptInput.tsx
morbargig: done

morbargig (commented):
please check now

morbargig (commented):
review

morbargig (commented):
add is-alive logic to show if the ai is working

PR #106, opened 2026-05-27
Add hash-based project deep links (FlowBolt frontend)

morbargig:
## Summary
FlowBolt builder `#/project/{id}` deep links — **separate from #96**.

**1 commit** · base: `main` (#99 ✅ #97 ✅ merged)

## Test plan
- [ ] Open `#/project/{id}` — correct project loads
- [ ] Refresh on hash URL
- [ ] Sidebar switch updates hash

### DIFF
--- a/frontend/src/constants/paths.ts
+++ b/frontend/src/constants/paths.ts
@@ -1,0 +1,9 @@
+/** Must stay in sync with backend flow44.paths constants. */
+
+export function previewBasePath(projectId: string): string {
+  return `/api/preview/${projectId}/proxy/`;
+}
+
+export function exportPublishedPath(projectId: string): string {
+  return `/api/export/${projectId}/published/`;
+}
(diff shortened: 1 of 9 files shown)


frontend/src/constants/paths.ts
```diff
@@ -0,0 +1,9 @@
+/** Must stay in sync with backend flow44.paths constants. */
```
88roy88, line 1: so better solution will be to have an endpoint to get the preview and export url... also ask @shalom-raichman this will work well with his custom-url feature

morbargig: i think if one merge the other will need to contact the other person to make sure the second merge is working,
he didn't merge it already right ?

88roy88: talk with him

morbargig: Shlomi is in a day off today, 
we talk and said that for today it ok if i can merge my changes and after that we can fix any issue together.

PR #109, opened 2026-06-02
Roym/sleep inactvie

88roy88:


docker-compose.yml
```diff
@@ -40,6 +40,7 @@ services:
       AIB_DB_PORT: *db-port
       AIB_DB_SCHEME: *db-scheme
       AIB_WORKSPACE_BASE_DIR: /var/lib/ai-builder/workspaces
+      AIB_SANDBOX_MODE: local
```
88roy88: check this.

docker-compose.yml
```diff
@@ -78,7 +79,7 @@ services:
     build: ./frontend
     user: root
     ports:
-      - "8888:80"
+      - "9888:80"
```
88roy88: revert

backend/Dockerfile
```diff
@@ -49,7 +49,9 @@ COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 # Create non-root user
 RUN useradd -m -u 1000 appuser \
     && mkdir -p /var/lib/flow-44/workspaces \
-    && chown -R appuser:appuser /var/lib/flow-44
+    && mkdir -p /var/lib/ai-builder/workspaces \
```
88roy88: check if ai-builder needed or stam old stuff

backend/src/flow44/api/chat.py
```diff
@@ -65,13 +65,8 @@ async def chat_ws(websocket: WebSocket, project_id: str) -> None:  # noqa: C901,
         return
 
     try:
-        sandbox = sandbox_manager.get_sandbox(project_id)
+        sandbox = await sandbox_manager.wake_sandbox(project_id)
```
88roy88: why wake before ensure?

backend/src/flow44/api/chat.py
```diff
@@ -65,13 +65,8 @@ async def chat_ws(websocket: WebSocket, project_id: str) -> None:  # noqa: C901,
...
         await sandbox_manager.ensure_ready(sandbox)
-    except SandboxNotFoundError:
```
88roy88, line 78: why we dont care about sandbox not found anymore?

backend/src/flow44/api/deps.py
```diff
@@ -18,8 +18,8 @@ def get_sandbox(project_id: str) -> PnpmSandbox:
 
 async def get_ws_sandbox(websocket: WebSocket, project_id: str) -> PnpmSandbox | None:
     try:
-        return sandbox_manager.get_sandbox(project_id)
-    except SandboxNotFoundError:
+        return await sandbox_manager.wake_sandbox(project_id)
+    except Exception:
```
88roy88: why?

backend/src/flow44/api/preview.py
```diff
@@ -37,21 +52,31 @@ async def get_preview_port(project_id: str, sandbox: Annotated[PnpmSandbox, Sand
...
-async def proxy_to_sandbox(  # noqa: E501
-    project_id: str, path: str, request: Request, sandbox: Annotated[PnpmSandbox, SandboxDep]
-) -> Response:
+async def proxy_to_sandbox(project_id: str, path: str, request: Request) -> Response:
     """Reverse proxy requests to the sandbox's dev server.
 
     Vite serves content under its ``base`` path, so the proxy forwards the
     full prefixed path to the upstream server.
+
+    If the sandbox was evicted (idle), this triggers re-creation and shows
```
88roy88: remove docstring (which is also wrong, we are not re-creating, just waking up)

docker-compose.yml
88roy88: Intentional for local dev — nsjail mode times out on pnpm install (60s limit kills the process). Production still uses namespaced mode via the Dockerfile ENV. This override is dev-only.

docker-compose.yml
88roy88: This is from the docker restructuring (removed nginx container, frontend serves directly). Will revert in a follow-up if we want to keep the nginx proxy setup.

backend/Dockerfile
88roy88: Needed — the docker-compose mounts the workspace volume to `/var/lib/ai-builder/workspaces` (line 42). Without this mkdir+chown, the volume mounts as root and the appuser can't write to it. The `/var/lib/flow-44` path is the default for non-Docker (local dev).

backend/src/flow44/api/chat.py
88roy88: Separation of concerns: `wake_sandbox` gets the sandbox object (allocates port, creates lightweight Python object). `ensure_dev_server` starts the dev server if not running. They're separate because terminal WebSocket only needs the object (PTY works without a dev server), while chat/preview need the server running. Merged them would block terminal on pnpm startup unnecessarily.

backend/src/flow44/api/chat.py
88roy88: `wake_sandbox` handles 'not found' internally — if the sandbox was evicted, it re-creates it. So `SandboxNotFoundError` can never escape this call. The remaining `except Exception` catches real infrastructure failures (port exhaustion, filesystem errors) which are properly logged.

backend/src/flow44/api/deps.py
88roy88: Changed from `get_sandbox` (raises if evicted) to `wake_sandbox` (re-creates if evicted). This is what allows terminal/errors/server-log WebSockets to wake sleeping projects on reconnect. The `except Exception` is now split: `SandboxNotFoundError` → close with 1008, other exceptions → logged + close with 1011.

backend/src/flow44/api/preview.py
88roy88: Done — removed the docstring.

backend/Dockerfile
88roy88: I think you're wrong and `/ai-builder` is the old stuff, just update whatever needed to flow44

docker-compose.yml
88roy88: move to 9999

docker-compose.yml
88roy88: I want you to check if its indeed doesnt work with nsjail mode, if we have the 60s issue, I want to see a log proofing it

backend/src/flow44/api/chat.py
88roy88: whats the point of wake_sandbox if we dont wait for it to be ready? do we have a use case for it?

backend/src/flow44/api/chat.py
88roy88: what if we got a wake-up on really non existing sandbox? (can it even happen?)

backend/src/flow44/api/deps.py
88roy88: I dont like general catch of Exception what are we trying to catch here?

docker-compose.yml
88roy88: Evidence from Docker logs: scaffold started at `17:06:53.247`, dev server started at `17:06:53.260` (13ms later). Zero `[scaffold]` log lines appeared (these are emitted per-line of pnpm install output). The `exec()` method in nsjail mode passes `--time_limit 60` (from `MAX_COMMAND_TIMEOUT=60`). On a cold pnpm store inside Docker, install takes 30-90s and gets killed by nsjail. The background dev server spawn (which uses `time_limit=0`) races against scaffold. Fixing the race is separate from the idle reaper — the `AIB_SANDBOX_MODE=local` override is a workaround for local dev only.

backend/src/flow44/api/chat.py
88roy88: Yes it can happen — if someone bookmarks a project URL then the project gets deleted from the DB. `wake_sandbox` would happily create a sandbox object (mkdir + port alloc) for a project that no longer exists. The protection is upstream: `chat_ws` checks `get_project(project_id)` before calling `wake_sandbox` (line 61-65). The terminal/errors/preview paths via `get_ws_sandbox` don't have that check — they'd create an orphan sandbox that gets reaped after 5 min of inactivity. Harmless but wasteful. We could add a DB existence check inside `wake_sandbox` if you want.

backend/src/flow44/api/deps.py
88roy88: Fixed — split it into two explicit catches:
```python
except SandboxNotFoundError:
    await websocket.close(code=1008, reason="No sandbox")
    return None
except Exception:
    logger.exception("Failed to wake sandbox for project %s", project_id)
    await websocket.close(code=1011, reason="Sandbox error")
    return None
```
The second catch handles: `RuntimeError` (port pool exhausted), `PermissionError`/`OSError` (filesystem issues during sandbox.start). These are logged now so we have visibility.

docker-compose.yml
88roy88: Done — moved to 9999.

backend/Dockerfile
88roy88: You're right — removed `/var/lib/ai-builder` from Dockerfile and updated docker-compose volume mount + env var to use `/var/lib/flow-44/workspaces`.

docker-compose.yml
88roy88: You were right — it's not a timeout issue. The actual failure is that the pnpm store directory (`/var/lib/flow-44/workspaces/.pnpm-store`) doesn't exist on first boot, so nsjail fails immediately trying to bind-mount it:

```
[W] mount('/var/lib/flow-44/workspaces/.pnpm-store' -> '/.pnpm-store') failed: No such file or directory
[E] initCloneNs(): Couldn't mount '/.pnpm-store'
[F] runChild(): Launching child process failed
```

Fixed by ensuring both `WORKSPACE_BASE_DIR` and `PNPM_STORE_DIR` are created at startup in `reconcile_workspaces`. Removing the `AIB_SANDBOX_MODE: local` override now.

PR #119, opened 2026-06-08
Make special user check work even if user has a suffix

yoni5353:


backend/pnpm-project-template/src/App.tsx
88roy88, line 10: can we have an is_special_user or something like that instead of repeating this logic?

88roy88: ignore me... we have in the frontend, this is the template///

88roy88 (approved):

backend/pnpm-project-template/src/App.tsx
```diff
@@ -1,6 +1,6 @@
 import { authConfig } from './auth/config';
 
-const SPECIAL_USERS = ['666royz', 'dev-user'];
+const SPECIAL_USERS = ['666royz', '619roym', '611noat', "dev-user"];
```
yoni5353, line 3: actually is this what you want in the resulted projects, or not? im having trouble following your whims

88roy88: this is fine, and you can add yourself as well :)

PR #108, opened 2026-06-02
Fix stale agent activity state

morbargig:
## Summary
- add backend agent activity tracking and alive endpoint
- reconcile frontend busy UI with backend alive state after reloads or websocket loss
- simplify agent phase selectors, progress rendering, and plan-approval handling

## Verification
- pnpm test src/stores/__tests__/chatHandlers.test.ts
- pnpm typecheck
- uv run pytest tests/api/test_iaagent_alive.py --no-cov

PR #122, opened 2026-06-09
fix: preserve agent progress cards on page refresh and WS reconnect

88roy88:
## Summary
- **Page refresh**: `finalizeHistoryReplayState` no longer clears replayed transient state — the alive poll handles cleanup if the agent is actually dead
- **WS reconnect**: `onClose` callback only fires on permanent disconnection, not transient auto-reconnects that resolve in seconds
- **Poll stability**: removed `agentAlive: false` from `TRANSIENT_RESET` so the poll doesn't die after `plan_overview` or other transient resets
- **UI consistency**: unified `EditorPanel` and `GlobalProgress` on `isAgentAlive` (poll-aware) instead of `isAgentWorking` (local-only)
- **Duplicate notification**: removed redundant `notifyAgentNeedsAttention` from phase handler (already handled by `plan_overview`)
- **Progress flicker**: progress bar shows 5% during the brief idle→first-phase gap instead of flickering to 0%

## Regression introduced by
PR #108 ("Fix stale agent activity state") — `a58b39a`

## Test plan
- [x] `npx vitest run` — all 112 tests pass
- [x] `npx tsc --noEmit` — clean
- [ ] Manual: start agent work → refresh page mid-execution → cards should persist
- [ ] Manual: start agent work → DevTools Network offline/online → cards should not vanish
- [ ] Manual: submit a plan → accept → progress bar tracks through execution without stalling

### DIFF
--- a/frontend/src/components/editor/EditorPanel.tsx
+++ b/frontend/src/components/editor/EditorPanel.tsx
@@ -7,7 +7,7 @@ loader.config({ monaco });
 import { Check, Loader2 } from 'lucide-react';
 import { useFilesStore } from '../../stores/files';
 import { useChatStore } from '../../stores/chat';
-import { isAgentWorking } from '../../stores/chatAgentState';
+import { isAgentAlive } from '../../stores/chatAgentState';
 import { useSessionStore } from '../../stores/session';
 import { WRITE_ROLES } from '../../types';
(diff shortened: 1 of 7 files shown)


morbargig (commented):
the code look fine, i trust you checked it and that it works :)

morbargig (approved):

PR #125, opened 2026-06-10
Fix data source run function for parameterless pacakges (missing POST method and headers)

yoni5353:


backend/src/flow44/ai/codegen/data_source_module.py
```diff
@@ -147,7 +147,12 @@ def _build_body(
     path = f"/api/data-source/{data_source_id}/run"
 
     if not required and not optional:
-        lines.append(f"  const res = await fetchWithAuth('{path}');\n")
+        lines.append(
+            f"  const res = await fetchWithAuth('{path}', {{\n"
+            f"    method: 'POST',\n"
```
88roy88: since its always post (also in the else case), lets just change fetchWithAuth to be post? or create a default wrapper for it something like fetchDataSource (for better AI understanding)

88roy88 (changes requested):

backend/src/flow44/ai/codegen/data_source_module.py
yoni5353: done

88roy88 (approved):

PR #126, opened 2026-06-10
Add data source generation to the FollowUp agent

### LINKED ISSUE #40: Follow up agent - Support for Data-Sources

yoni5353:
Closes https://github.com/88roy88/FlowBolt/issues/40

extracts the package anaylze function to a dedicated file and uses it in both agents; alembic revisions explained in the comments

this also adds a "new" icon to newly generated files in the chat file diff


backend/src/flow44/db/project.py
```diff
@@ -111,20 +97,7 @@ async def get_project_data_sources(project_id: str) -> list[dict[str, Any]]:
         project = await session.get(Project, project_id)
         if project is None:
             return []
-        if project.data_sources:
-            return project.data_sources
-        # TODO: what is this VVVVV
```
yoni5353, line 115: lol, these were old fields in the schema that are not used atm, and were changed to the `data_sources` field that saves a list of data sources instead of just one.

i removed them from the schema, so correct me if im wrong

backend/src/flow44/db/project.py
```diff
@@ -85,22 +84,15 @@ async def update_project_model(project_id: str, model: str) -> None:
...
-            project.data_source_context = data_source_context
-            project.updated_at = datetime.now(UTC).isoformat()
-            session.add(project)
-            await session.commit()
+_TRANSIENT_KEYS = {"sample_data", "generated_files"}
 
 
-async def update_project_data_sources(project_id: str, data_sources: list[dict[str, Any]]) -> None:
+async def update_project_data_sources(project_id: str, data_sources: Sequence[DataSourceContext]) -> None:
+    stored = [{k: v for k, v in ds.items() if k not in _TRANSIENT_KEYS} for ds in data_sources]
```
yoni5353: it's critical to omit `sample_data` here as it can be very heavy (dozens of megabytes). I also included an alembic revision to delete this field from the prod db

unless you think it is important to save? i think it's probably redundant after the ai analyzes the data source anyway

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -107,12 +114,12 @@ async def read_file(path: str, offset: int = 0, limit: int = MAX_READ_LINES) ->
         @tool
         async def write_file(path: str, content: str) -> str:
             """Write the full content of a file, creating it if needed. For small changes, prefer edit_file."""
-            import difflib  # noqa: PLC0415
```
88roy88, line 112: tnx 👑

backend/src/flow44/ai/agents/followup/templates/followup.jinja2
```diff
@@ -1,5 +1,19 @@
...
+The following data sources were just attached and their TypeScript modules have been generated and written to the sandbox. Pre-built async API functions and TypeScript types have ALREADY been generated for each one. You MUST use these existing files. Do NOT create new fetch logic or types for data sources — they are ready to import.
+{% for ds in new_data_source_contexts %}
+### Data Source: {{ ds.data_source_name }} (ID: {{ ds.data_source_id }}) — `{{ ds.module_path }}`
+- **Data Schema:** {{ ds.data_schema }}
+- **Relevant Fields:** {{ ds.relevant_fields }}
+- **Data Characteristics:** {{ ds.data_characteristics }}
+
+Usage: `import { dataSource{{ ds.sanitized_name }} } from './dataSources/{{ ds.sanitized_name }}';`
+{% endfor %}
+Use these modules when wiring up the new data sources into the app.
```
88roy88: I would be more explicit about that.
something in the line of:
"if the user asked you to to add or replace .... blah blah blah"

help the model to understand what it should do with this

88roy88 (changes requested):

88roy88:
See my PR adding to this one, (#131) and see the comment about splitting the datasources to another table in (#129) I think this is the way to go to simplify a lot of things both here and there

yoni5353 (commented):
oh i forgot to submit these comments

yoni5353:
> See my PR adding to this one, (#131) and see the comment about splitting the datasources to another table in (#129) I think this is the way to go to simplify a lot of things both here and there

yeah, i agree. i created this new table `project_data_sources` in this branch and moved the related functions there. 

gonna revert related changes in #129

backend/src/flow44/db/project_data_source.py
```diff
@@ -0,0 +1,86 @@
...
+    data_source_id: str = ""
+    data_source_name: str = ""
+    sanitized_name: str = ""
+    queries: list[dict[str, Any]] = Field(default_factory=list)
+    params_info: dict[str, Any] = Field(default_factory=dict)
+    can_run_without_input: bool = False
+    data_schema: str = ""
+    relevant_fields: str = ""
+    data_characteristics: str = ""
+    integration_notes: str = ""
```
yoni5353: i continued the idea where the DataSourceContext extends the DataSource which i think makes sense.

This type duplication hurt my soul a little but i don't know of a better way to do this with sqlmodel. At the very least it has the util to convert between them, and they are grouped in the same file

backend/src/flow44/db/project_data_source.py
```diff
@@ -0,0 +1,86 @@
...
+    data_source_name: str = Field(default="")
+    sanitized_name: str = Field(default="")
+    queries: list[dict[str, Any]] = Field(default_factory=list, sa_column=Column(JSON, default=[]))
+    params_info: dict[str, Any] = Field(default_factory=dict, sa_column=Column(JSON, default={}))
+    can_run_without_input: bool = Field(default=False)
+    data_schema: str = Field(default="")
+    relevant_fields: str = Field(default="")
+    data_characteristics: str = Field(default="")
+    integration_notes: str = Field(default="")
+    param_ux_hints: str = Field(default="")
```
yoni5353: @88roy88 
Question: do you think we need to remove these fields from the db (especially since you keep them in them markdown in your branch?) i can see UI use cases for keeping stuff like `queries`, but we could scrap the ui's analysis, i suppose

for now i just kept all fields for the heck of it, they are not heavy anyway

backend/src/flow44/ai/agents/followup/templates/followup.jinja2
yoni5353: i reworked this sentence. is it too elaborate now? let me know

backend/src/alembic/versions/2026_06_14_1144-640c819eb8c0_add_project_data_sources_table.py
```diff
@@ -0,0 +1,93 @@
+"""add_project_data_sources_table
```
88roy88: given this alembic, do we need the 2 previous ones?

backend/src/flow44/db/project_data_source.py
88roy88: lets keep them for now

backend/src/flow44/db/project_data_source.py
88roy88: The idiomatic SQLModel move here is a shared non-table base. SQLModel's whole "Base / table / Public / Create" pattern exists precisely for this — you define the common fields once on a plain SQLModel subclass (no table=True), then the table model and the in-memory model both inherit it and add only what's unique to them.
```python
class DataSourceBase(SQLModel):
    project_id: str = ""
    type: str = "flow_package"
    data_source_id: str = ""
    data_source_name: str = ""
    sanitized_name: str = ""
    queries: list[dict[str, Any]] = Field(default_factory=list, sa_type=JSON)
    params_info: dict[str, Any] = Field(default_factory=dict, sa_type=JSON)
    can_run_without_input: bool = False
    data_schema: str = ""
    relevant_fields: str = ""
    data_characteristics: str = ""
    integration_notes: str = ""
    param_ux_hints: str = ""


class ProjectDataSource(DataSourceBase, table=True):
    __tablename__ = "project_data_sources"
    __table_args__ = (
        UniqueConstraint("project_id", "type", "data_source_id", name="uq_project_data_source"),
    )
    id: str = Field(default_factory=lambda: str(uuid.uuid4()), primary_key=True)
    project_id: str = Field(index=True)  # re-declare just to add the index


class DataSourceContext(DataSourceBase):
    id: str = ""
    sample_data: dict[str, Any] | None = None
    module_path: str = ""
    generated_files: dict[str, str] = Field(default_factory=dict)
```

88roy88 (changes requested):

backend/src/alembic/versions/2026_06_14_1144-640c819eb8c0_add_project_data_sources_table.py
yoni5353: i combined all three

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -173,9 +178,13 @@ async def edit_file(path: str, search: str, replace: str) -> str:
         return ToolExecutor([grep, glob, read_file, write_file, edit_file])
 
     @observe(name="followup-agent-run")  # type: ignore[untyped-decorator]
-    async def run(self, content: str) -> None:
+    async def run(self, content: str, data_source_ids: list[str] | None = None) -> None:
         self._setup_trace(["follow-up-agent"])
 
+        new_data_source_contexts: list[DataSourceContext] = []
+        if data_source_ids:
```
88roy88, line 187: so we are not really check if they are new? if they provide me a package-id that is already implemented?

backend/src/flow44/db/project_data_source.py
```diff
@@ -0,0 +1,76 @@
...
+
+from sqlalchemy import JSON, UniqueConstraint
+from sqlmodel import Field, SQLModel, col, select
+
+from flow44.db import database
+
+
+class DataSourceBase(SQLModel):
+    project_id: str = ""
+    type: str = "flow_package"
```
88roy88, line 13: as best practice, I like literals or enums for stuff like this. for now I can live with it as its only one value, but its kinda stinks

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -209,14 +219,79 @@ async def run(self, content: str) -> None:
...
+                    for sid in data_source_ids
+                ]
+            )
+        except Exception:
+            await self.emit({"type": "error", "message": "Failed to fetch required data source data."})
+            raise
+
+        contexts: list[DataSourceContext] = list(results)
+
+        for ctx in contexts:
```
88roy88: can we split this function to not be this huge?

backend/src/flow44/ai/agents/followup/agent.py
yoni5353: if i reattach an existing package, it overrides the module with the updated fetch which is probably good if new parameters and cubes were added, which sounds quite common to me. so we probably want to GET the package anyway

maybe ill just adjust the prompt so it will keep in mind it can be the same package? from my testing it handled it well as-is though, so i think its fine unless you think otherwise

backend/src/flow44/db/project_data_source.py
yoni5353: sure

backend/src/flow44/ai/agents/followup/agent.py
yoni5353: i split it to two, and improved the asyncio to also generate module immediately after each fetch

backend/src/flow44/ai/agents/followup/agent.py
88roy88: I agree this is good for updating the package, but we are not telling it to fix all the places which depend on it, which can break things.

88roy88: also the prompt text refer to them as new...

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -209,14 +219,79 @@ async def run(self, content: str) -> None:
...
         # TODO: do we need both events?
         await self.emit({"type": "phase", "phase": "complete"})
         await self.emit({"type": "action_complete"})
 
+    async def _fetch_and_generate_new_data_sources(
+        self, user_content: str, data_source_ids: list[str]
+    ) -> list[DataSourceContext]:
+        await self.emit({"type": "phase", "phase": "fetching_data_sources"})
+
+        async def _fetch_and_write(sid: str) -> DataSourceContext:
```
88roy88: can we not make inline functions :) just move it out bro

backend/src/flow44/ai/agents/followup/agent.py
yoni5353: so my fix will be -- fix the prompt to notice if the package already exists and fix the integration of it

correct?

or do you think i should manually check the id and add a special prompt for that?

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -209,14 +219,79 @@ async def run(self, content: str) -> None:
...
+        )
+        await self.emit({"type": "file", "path": module_path, "content": content})
+        if diff_str:
+            self._diffs.append(FileDiff(path=module_path, diff=diff_str, is_new=is_new_module))
+        if module_path not in self._files_changed:
+            self._files_changed.append(module_path)
+
+    async def _persist_data_sources(self, new_contexts: list[DataSourceContext]) -> None:
+        """Merge newly attached data sources into the project's stored list (by id)."""
+        existing = await get_project_data_sources(self.project_id)
```
88roy88: so we do have the logic for finding if its really new or not ;)

backend/src/flow44/ai/agents/followup/agent.py
88roy88: it can't be only in prompt level, as by the time the prompt is running, we already generated (and override) the files

backend/src/flow44/ai/agents/followup/agent.py
yoni5353: caught red handed :o

backend/src/flow44/ai/agents/followup/agent.py
yoni5353: done

didn't realize u don't like those, i like them nested but will keep that in mind for you -.-

backend/src/flow44/ai/agents/followup/agent.py
yoni5353: ok i added said logic as a new exiting_data_sources_context and added a section in the prompt. tell me what you think

88roy88 (approved):

PR #129, opened 2026-06-11
Improve preformance or projects.py routes

yoni5353:
make the projects GET and DELETE routes faster. (omitting the large `data_sources` field from listings, and adjusting deletion so that deleting the sandbox is an afterthought and non-blocking)

### DIFF
--- a/backend/src/flow44/api/projects.py
+++ b/backend/src/flow44/api/projects.py
@@ -56,12 +56,14 @@ async def list_user_projects(user_id: UserDep) -> list[dict[str, Any]]:
     user_perms = get_admin_permissions() if is_admin(user_id) else set()
     can_read_all = has_permission(user_perms, Permission.read)
 
+    exclude = {"data_sources"}
+
     if can_read_all:
         all_projects = await list_all_projects()
         result: list[dict[str, Any]] = []
         for p in all_projects:
             role = "owner" if p.user_id == user_id else "admin"
-            result.append(p.model_dump() | {"role": role})
+            result.append(p.model_dump(exclude=exclude) | {"role": role})
         return result
 
     owned = await db_list_user_projects(user_id)
@@ -69,9 +71,9 @@ async def list_user_projects(user_id: UserDep) -> list[dict[str, Any]]:
 
     result = []
     for p in owned:
-        result.append(p.model_dump() | {"role": "owner"})
+        result.append(p.model_dump(exclude=exclude) | {"role": "owner"})
     for p, role in shared:
-        result.append(p.model_dump() | {"role": role})
+        result.append(p.model_dump(exclude=exclude) | {"role": role})
 
     return result
 
--- a/backend/src/flow44/api/projects.py
+++ b/backend/src/flow44/api/projects.py
@@ -66,8 +66,7 @@ async def list_user_projects(user_id: UserDep) -> list[dict[str, Any]]:
             result.append(p.model_dump(exclude=exclude) | {"role": role})
         return result
 
-    owned = await db_list_user_projects(user_id)
-    shared = await list_shared_projects(user_id)
+    owned, shared = await asyncio.gather(db_list_user_projects(user_id), list_shared_projects(user_id))
 
     result = []
     for p in owned:
--- a/backend/src/flow44/api/projects.py
+++ b/backend/src/flow44/api/projects.py
@@ -6,7 +6,7 @@ import asyncio
 import logging
 from typing import Any
 
-from fastapi import APIRouter, HTTPException
+from fastapi import APIRouter, BackgroundTasks, HTTPException
 from pydantic import BaseModel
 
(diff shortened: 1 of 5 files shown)


88roy88:
dont you think that a better approach would be to just move the datasource out of this table to another one (could help later with all the stuff we talked about with other flow integrations)

frontend/src/types/index.ts
```diff
@@ -75,6 +75,9 @@ export interface Project {
   selected_model?: string;
   published_url?: string;
   role?: ProjectRole;
+
+  /** Not included in list responses, fetch individually to access */
+  data_sources?: string;
```
88roy88: they are not string though...

backend/src/flow44/api/projects.py
```diff
@@ -56,22 +56,23 @@ async def list_user_projects(user_id: UserDep) -> list[dict[str, Any]]:
...
         result: list[dict[str, Any]] = []
         for p in all_projects:
             role = "owner" if p.user_id == user_id else "admin"
-            result.append(p.model_dump() | {"role": role})
+            result.append(p.model_dump(exclude=exclude) | {"role": role})
         return result
 
-    owned = await db_list_user_projects(user_id)
-    shared = await list_shared_projects(user_id)
+    owned, shared = await asyncio.gather(db_list_user_projects(user_id), list_shared_projects(user_id))
```
88roy88, line 69: nice

backend/src/flow44/api/projects.py
```diff
@@ -119,11 +120,12 @@ async def update_project_selected_model(
 @router.delete("/{project_id}", status_code=204)
 async def delete_existing_project(
     project: ProjectDep,
+    background_tasks: BackgroundTasks,
     _perms: set[Permission] = require_permission(Permission.delete),
 ) -> None:
-    await sandbox_manager.destroy_sandbox(project.id)
     idle_reaper.remove(project.id)
     await delete_project(project.id)
+    background_tasks.add_task(sandbox_manager.destroy_sandbox, project.id)
```
88roy88, line 128: nice, when working with bg job, its bets to add some logs in case of something dies.

backend/src/flow44/db/project.py
```diff
@@ -42,15 +43,20 @@ async def get_project(project_id: str) -> Project | None:
 
 async def list_user_projects(user_id: str) -> list[Project]:
     async with database.async_session() as session:
-        query = select(Project).where(Project.user_id == user_id).order_by(col(Project.created_at).desc())
+        query = (
+            select(Project)
+            .where(Project.user_id == user_id)
+            .order_by(col(Project.created_at).desc())
+            .options(defer(Project.data_sources))  # type: ignore[arg-type]
```
88roy88: check if using `col` like we did in the other places solves the type ignore

backend/src/flow44/db/project.py
```diff
@@ -42,15 +43,20 @@ async def get_project(project_id: str) -> Project | None:
...
+        )
         result = await session.execute(query)
         return list(result.scalars().all())
 
 
 async def list_all_projects() -> list[Project]:
     """System-level: returns every project across all users. Never call from a request handler."""
```
88roy88: `Never call from a request handler` is not true I think :)

88roy88 (changes requested):

backend/src/flow44/api/projects.py
yoni5353: it does log on failure inside `destroy_sandbox`, is that enough or did you mean i should log always?

backend/src/flow44/api/projects.py
yoni5353: : )

backend/src/flow44/db/project.py
yoni5353: removed that part

backend/src/flow44/db/project.py
yoni5353: the AI gods are telling me it wouldnt have helped, but i removed this in any case

frontend/src/types/index.ts
yoni5353: true, mb, i removed this line altogether

yoni5353:
> dont you think that a better approach would be to just move the datasource out of this table to another one (could help later with all the stuff we talked about with other flow integrations)

I reverted all things related to the `data_sources` filed, as i did move them to another table in #126. 

the other changes are still worthwhile, imo

backend/src/flow44/db/project.py
```diff
@@ -48,7 +48,7 @@ async def list_user_projects(user_id: str) -> list[Project]:
 
 
 async def list_all_projects() -> list[Project]:
-    """System-level: returns every project across all users. Never call from a request handler."""
+    """Returns every project across all users."""
```
88roy88, line 51: lets just remove the docstring all together :)

88roy88 (approved):

PR #130, opened 2026-06-11
feat: fix followup agent memory and datasource support

88roy88:
## Summary
- Followup agent now saves assistant responses (with compact tool-call log + files-changed) to `chat_messages`, fixing broken conversation memory
- Datasource context is now passed to the followup system prompt — both existing project datasources and newly attached ones
- Introduces `ChatAgent` base class shared by FollowUpAgent and FixErrorAgent
- Extracts `build_data_source_context()` as shared utility used by plan + followup agents

## Test plan
- [ ] Send a followup message → verify `chat_messages` has assistant row with tool log + answer
- [ ] Send another followup → confirm LLM sees interleaved user/assistant history
- [ ] Attach a datasource to a followup → confirm system prompt includes DS schema
- [ ] Trigger fix_error → confirm it also saves response
- [ ] Verify initial plan with datasources still works after refactor

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #131, opened 2026-06-11
feat: add conversation memory to followup/fix agents

88roy88:
## Summary
Builds on top of #126 (yonis/package-followup) to add:
- **ChatAgent** base class shared by FollowUpAgent and FixErrorAgent
- **Assistant responses saved to chat_messages** — includes compact **tool-call** log and llm **reasoning** + files-changed summary, fixing missing conversation memory
- **Existing project datasources** loaded into the followup system prompt (alongside Yoni's new-DS handling)
- **Parallelized** `_build_context` and `get_messages` for lower latency

## Test plan
- [ ] Send a followup → verify `chat_messages` has assistant row with tool log + answer
- [ ] Send another followup → confirm LLM sees interleaved user/assistant history
- [ ] Verify existing project DS appear in followup prompt
- [ ] Trigger fix_error → confirm it also saves response

Supersedes #130


backend/src/flow44/ai/agents/followup/templates/followup.jinja2
```diff
@@ -20,6 +20,15 @@ Use these modules when wiring up the new data sources into the app.
 ## File Tree
 {{ file_tree }}
 
+{% if existing_data_source_contexts %}
```
88roy88: lets make this part clearer by telling the ai to *only( read the doc files *if needed*

backend/src/flow44/ai/agents/fix_error/agent.py
```diff
@@ -248,6 +249,16 @@ async def _step_retry(self, state: FixErrorState) -> FixErrorState:
 
     async def _step_complete(self, state: FixErrorState) -> FixErrorState:
         """Step: Complete the fix process."""
+        files = [p for p, _ in state.generated_files]
+        steps = [
+            {
+                "tool": "fix_error",
+                "args": {"file": state.error_file or "unknown"},
+                "resultPreview": f"fixed {len(files)} file(s)",
```
88roy88: I dont like that we are in typescript camelCase here and not A. with a basemodel for it, and B. in snake case, let the frontend handle this...

backend/src/flow44/ai/agents/_chat_agent.py
```diff
@@ -0,0 +1,54 @@
...
+
+        for step in steps:
+            tool = step.get("tool", "?")
+            args = step.get("args", {})
+            preview = step.get("resultPreview", "")
+
+            args_str = " ".join(
+                f"{k}={v!r}" if not isinstance(v, str) else f"{k}={v}" for k, v in args.items() if k != "content"
+            )
+            result_short = preview[:80].replace("\n", " ").strip()
```
88roy88, line 27: this is fine, but I thought we can also add to the tool class an option to return a custom short preview that we can use here. for example for grep something like: `found X matches` or whatever....

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -314,7 +316,9 @@ async def _build_context(self) -> dict[str, str]:
         except Exception:
             file_tree = "(unable to list files)"
 
-        return {"summary": summary, "file_tree": file_tree}
+        existing_ds = (project.data_sources or []) if project else []
```
88roy88: I still think we need to make this lazy load and or move it to another table..

yoni5353: yeah, i created the new table on the base branch, so if you agree with the change this value and its type need to be adjusted

(can do it myself if ill have confirmation that you're ok with my table there)

yoni5353 (commented):
looks good, funny that it didn't save the messages till now. i learned a lot

so just awaiting thepossible change to `data_sources` in #126

backend/src/flow44/ai/agents/followup/templates/followup.jinja2
yoni5353: merged that with the other prompt

backend/src/flow44/ai/agents/fix_error/agent.py
yoni5353: fixed

backend/src/flow44/ai/agents/_chat_agent.py
yoni5353: i added an optional `short_preview` for this. (it's this commit [83f03ae])(https://github.com/88roy88/FlowBolt/pull/131/commits/83f03ae89b8b68e749692a7c1eb6c26a4a505821)

it also felt like it's specifically handy for `read_file` because the regular preview always returns the beginning of the file "import {} ...." which is redundant for context

backend/src/flow44/ai/agents/followup/agent.py
yoni5353: fixed in merge

yoni5353:
I'm still stumped on how best to solve the thing where it tries to match the generated message to how it is saved in the history.

changing the history format did not work, and adding a prompt that specifically tells it to not do so didn't work either

maybe lets talk about it IRL unless u have a quick fix

backend/src/flow44/ai/agents/fix_error/agent.py
88roy88, line 168: btw, Neta asked if we could have the diff showing in the front like we have in the follow up agent. we can open another ticket for that and handle it separately

backend/src/flow44/ai/agents/followup/templates/followup.jinja2
```diff
@@ -20,13 +20,15 @@ Based on the user's request, determine their intent:
 ## Existing Data Sources
 The following data sources are already integrated in the app. Their TypeScript modules have been regenerated — if the function signature or return type changed, update the relevant call sites to match.
 Based on the user's request, you may also need to build new UI around them (new section, option, or replacement).
+If you need to compare against the previous version of the schema or parameter details, use `read_file` on the `.docs.md` file.
```
88roy88: yes? arent we update(override) those as well? I think he never really gets to see the old stuff

yoni5353: yeah you're right, i got the order mixed up in my head somehow. deleting this.

backend/src/flow44/ai/agents/fix_error/agent.py
yoni5353: yeah i noticed that too, ill do that

yoni5353: https://github.com/88roy88/FlowBolt/pull/157

yoni5353:
@88roy88 github doesn't let me add you as a reviewer but this branch is ready for you again

backend/src/alembic/versions/2026_07_01_2132-1fd0c176782a_add_tool_call_and_tool_result_values_to_.py
```diff
@@ -0,0 +1,38 @@
+"""Add tool_call and tool_result values to chatrole enum
```
88roy88: I have this feeling that you didnt used the auto-generate for it? 
can you use the `alembic revision --auto-generate` as base for your migration and edit if needed

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -89,10 +90,11 @@ async def glob(pattern: str = "*") -> str:
             results = await sandbox.glob(pattern)
             if not results:
                 return "No files found matching pattern."
-            return "\n".join(results)
+            value = "\n".join(results)
+            return ToolResult(value=value, short_preview=f"Found {len(results)} file(s).")
 
         @tool
-        async def read_file(path: str, offset: int = 0, limit: int = MAX_READ_LINES) -> str:
+        async def read_file(path: str, offset: int = 0, limit: int = MAX_READ_LINES) -> str | ToolResult:
```
88roy88: lets maybe create a type for when tool want to return with short_preview but its not the full ToolResult?
wdyt?

88roy88: also this is not `str |` as its always the new type

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -230,6 +234,8 @@ async def run(self, content: str, data_source_ids: list[str] | None = None) -> N
         if answer:
             await self.emit({"type": "text", "content": answer})
 
+        await self._save_response(answer or "", self._steps)
```
88roy88, line 237: talk to me about this, I think one of us is missing something.

README.md
```diff
@@ -21,7 +21,10 @@ AI-powered web app builder. The backend is a FastAPI service; the frontend is a
 
 # Apply all pending migrations
 
-`uv run alembic upgrade head`
+```bash
```
88roy88: add also the autogenerate command :)

backend/src/flow44/ai/agents/followup/agent.py
88roy88: I think we are missing the assistant messages that are not tool calling, I might be wrong and miss something here.

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -291,8 +296,7 @@ async def _persist_data_sources(
             by_id[(ctx.type, ctx.data_source_id)] = ctx
         await update_project_data_sources(self.project_id, list(by_id.values()))
 
-    # TODO: We will want to have a smarted memory system in the future
```
88roy88, line 294: can you check langfuse to see how fast our memory is growing now after this upgrade?

backend/src/flow44/ai/agents/_chat_agent.py
```diff
@@ -0,0 +1,29 @@
...
+
+class ChatAgent(BaseAgent):
+    """Base for agents that participate in the chat history (followup, fix_error)."""
+
+    async def _save_response(
+        self,
+        answer: str,
+        steps: list[dict[str, Any]],
+    ) -> None:
+        for step in steps:
```
88roy88, line 15: yes, I think we are missing the assistant non final messages here
(i.e "sure im going to call the grep tool first <tool call> <tool result> oh great now ill call the read file <tool call> .... ok here is my final answer (this is saved)")

backend/src/flow44/ai/agents/analyze_data_source.py
```diff
@@ -82,4 +84,71 @@ def generate_data_source_files(ctx: DataSourceContext) -> dict[str, str]:
...
+        if isinstance(obj, list):
+            return [_redact(obj[0])] if obj else []
+        if obj is None:
+            return None
+        return _TYPE_PLACEHOLDERS.get(type(obj), "<string>")
+
+    return json.dumps(_redact(sample), indent=2)
+
+
+def _generate_data_source_docs(
```
88roy88, line 120: what do you think about moving to use jinja for those templates as well? its so ugly to keep building it here in python

88roy88:
@yoni5353 back at ya

backend/src/alembic/versions/2026_07_01_2132-1fd0c176782a_add_tool_call_and_tool_result_values_to_.py
yoni5353: apparently `--autogenerate` does not detect enum values changes D: (at least by default. requires another dependency)

so i used the command for the format but the content looks the same still

backend/src/flow44/ai/agents/followup/agent.py
yoni5353: i mean isn't that just `ToolResult`? it only needs a message so it's already quite thin, just `ToolResult(value=,short_preview=)`, which is what i added where it felt needed

yoni5353: actually the `try` could return a string here, but i changed it to use a ToolResult too

obviously i considered changing everything to `ToolResult`, but it seemed out of scope for this branch, and a little verbose like u said above, so i just changed to it where i needed

README.md
yoni5353: it actually already existed below so i added a link and made this a proper step

backend/src/flow44/ai/agents/_chat_agent.py
yoni5353: hmm, that's true, but it looks like we're not saving his reasoning at all atm, and not displaying anything in the ui either! (so technically it's not a message...)

what do you think we should? add the reasoning to the chat_messages and not agent_events? add to both? 

show it in the ui? i think that'd be great, just seems a little verbose

yoni5353: in any case i think we can do this on a different branch unless you think othewise, just because its a signifcant change

backend/src/flow44/ai/agents/analyze_data_source.py
yoni5353: sure, done

backend/src/flow44/ai/agents/followup/agent.py
yoni5353: i eyeballed it a little bit but i dont really know what im supposed to learn. it's doesn't make a huge different because it's not that much text for now. 

obviously for long sessions it will be heavy, do you want to set some limit on messages or we'll keep it for later like u said?

yoni5353:
@88roy88 pong

backend/src/flow44/ai/agents/_chat_agent.py
88roy88: I think we do show the ai non final messages now, dont we?

backend/src/alembic/versions/2026_07_01_2132-1fd0c176782a_add_tool_call_and_tool_result_values_to_.py
88roy88: I love the extra dependency :) I'm so used to it that I forgot its an addon. Let's use it

yoni5353: done!

backend/src/flow44/ai/agents/_chat_agent.py
yoni5353: ok, i so i saved this so-called `reasoning_content` that the ai emits into a separate message type `reasoning`. played around with it, looks to be working fine

yoni5353:
@88roy88 pong, now i also save the reasoning to the db

yoni5353 (approved):

PR #135, opened 2026-06-14
PR 1: Configure pnpm 11 project installs

morbargig:
## Summary
- update frontend CI to pnpm 11
- require pnpm 11 for frontend, mocks, and the generated-app template
- approve esbuild install scripts for pnpm 11
- clean and install mock/template dependencies from `make install`

## Validation
- `git diff --cached --check`
- frontend `pnpm install --frozen-lockfile --offline`
- package JSON and YAML parsing

Replaces #134 after renaming the source branch.


backend/pnpm-project-template/package.json
```diff
@@ -21,5 +21,8 @@
     "tailwindcss": "^3.4.1",
     "typescript": "^5.2.2",
     "vite": "^5.1.4"
+  },
+  "engines": {
```
88roy88, line 25: why do we want to limit the pnpm version? why not keep it open like before?

backend/example.env
```diff
@@ -93,7 +93,7 @@ UV_INSECURE_HOST2=
 
 # --- Frontend Build & Runtime Configuration ---
 # Used as --build-arg in frontend/Dockerfile
-FRONTEND_BASE_IMAGE=node:20-slim
+FRONTEND_BASE_IMAGE=node:24-slim
```
88roy88, line 96: you checked it in NH, right? also in Openshift?

Makefile
```diff
@@ -46,7 +46,8 @@ dev-mocks: kill-port-$(DEV_PORT_MOCKS)
 # Install dependencies (also installs Husky git hooks)
 install:
 	cd frontend && pnpm install
-	cd mocks/flapi-mock && pnpm install
+	cd mocks/flapi-mock && pnpm clean --lockfile && pnpm install
```
88roy88, line 49: why are we `pnpm clean --lockfile`?

88roy88 (changes requested):

backend/pnpm-project-template/package.json
morbargig: to force the team and the environment to be at the same tech.
and to be updated.

backend/example.env
morbargig: i know it's should works, when i was using docker images on my project they was updated 24 is node lts

morbargig: and we can change it, it's the env still

Makefile
morbargig: because it took me more than 15/20 minutes to install the project, and this fix some use case that the install crush.

the lock on mac/linux/windows is not the same sometimes

morbargig (commented):
review

88roy88 (approved):

PR #137, opened 2026-06-14
PR 2: Add generated app routing and shared publishing

morbargig:
## Summary
- add React Router as an optional generated-app package with basename-safe prompt rules
- configure preview and published builds with public base paths
- publish complete Vite dist directories and serve shared assets with SPA fallback
- add routing, template, shared-serving, and publishing regression tests

## Stack
- depends on #138
- excludes ReGraph, Leaflet, table packages, lockfile churn, and local Docker changes

## Validation
- focused routing and publishing tests: 87 passed
- Ruff checks passed
- `git diff --cached --check`


88roy88:
why this one is on top of the AI one? why not just from develop? is it really dependent?

PR #136, opened 2026-06-14
PR 2: Add optional package support for generated apps

morbargig:
## Summary
- add a whitelist-driven optional package selection and installation flow
- support MUI and Recharts for generated apps
- enforce generated-file and dependency contracts across execute, fix-error, and follow-up flows
- add package-specific planning, code generation, and repair prompt templates

## Stack
- depends on #135
- excludes React Router, maps, tables, ReGraph, preview routing, and shared publishing

## Validation
- `git diff --cached --check`
- focused optional-package tests: 46 passed
- Ruff checks passed

PR #138, opened 2026-06-15
PR 1: Add protected file contract for generated apps

morbargig:
## Summary
- add a backend-enforced generated app file safety contract
- block AI writes to protected project files, unsafe paths, protected source files, and protected source folders
- render file-safety prompt rules from the same contract values used by backend enforcement
- move the shared file-safety prompt partial out of `execute/templates` into agent-level shared templates
- apply the contract across execute, fix-error, follow-up, and architecture prompts
- type prompt safety context and make file-safety template context explicit at render call sites

## Stack
- PR 1 of 3
- base: develop
- followed by optional packages in PR #143 and routing/shared publishing in PR #142

## Validation
- `uv run ruff check src/flow44/ai/generated_app_contract.py src/flow44/ai/agents/plan/prompts.py src/flow44/ai/agents/followup/prompts.py src/flow44/ai/agents/execute/prompts.py src/flow44/ai/agents/fix_error/prompts.py`: passed
- `UV_CACHE_DIR=/tmp/flowbolt-uv-cache uv run mypy --config-file mypy.toml src/flow44/ai/generated_app_contract.py src/flow44/ai/agents/plan/prompts.py src/flow44/ai/agents/followup/prompts.py src/flow44/ai/agents/execute/prompts.py src/flow44/ai/agents/fix_error/prompts.py`: passed
- `uv run pytest tests/ai/test_prompts.py tests/ai/test_generated_app_contract.py --no-cov`: 29 passed


backend/src/flow44/ai/generated_app_contract.py
```diff
@@ -0,0 +1,121 @@
...
+
+import re
+from pathlib import PurePosixPath
+
+
+class GeneratedAppContractError(ValueError):
+    """Raised when generated output violates the app-generation contract."""
+
+
+_PROTECTED_FILES = {
```
morbargig: Copied from @88roy88 on PR #136: I dont know if im fully in on the idea of protected files, its starting to grow on me though.
anyway, I would manage it differently, with per file marker inside the file and protection in the tool level

Original: https://github.com/88roy88/FlowBolt/pull/136#discussion_r3412842399

backend/src/flow44/sandbox/pnpm_mixin.py
```diff
@@ -42,6 +45,31 @@ async def scaffold(self, template_dir: str) -> None:
...
+            return False
+        with open(pkg_path, encoding="utf-8") as handle:
+            pkg = json.load(handle)
+        return package_name in pkg.get("dependencies", {})
+
+    def _npm_package_available(self, package_name: str) -> bool:
+        module_path = os.path.join(self.workspace_dir, "node_modules", *package_name.split("/"))
+        return self._npm_package_declared(package_name) and os.path.exists(module_path)
+
+    async def enable_optional_packages(self, package_names: list[str]) -> None:
```
morbargig: Copied from @88roy88 on PR #136: this is not enable, this is install

Original: https://github.com/88roy88/FlowBolt/pull/136#discussion_r3412843938

backend/src/flow44/sandbox/pnpm_mixin.py
```diff
@@ -42,6 +45,31 @@ async def scaffold(self, template_dir: str) -> None:
...
 
+    def _npm_package_declared(self, package_name: str) -> bool:
+        pkg_path = os.path.join(self.workspace_dir, "package.json")
+        if not os.path.isfile(pkg_path):
+            return False
+        with open(pkg_path, encoding="utf-8") as handle:
+            pkg = json.load(handle)
+        return package_name in pkg.get("dependencies", {})
+
+    def _npm_package_available(self, package_name: str) -> bool:
```
morbargig: Copied from @88roy88 on PR #136: not a fan of this logic.
the logic I would expect is per project save in db the required packages, and to have call add_package (Which will call pnpm add that is also installing) right at the agent level

Original: https://github.com/88roy88/FlowBolt/pull/136#discussion_r3412856141

backend/src/flow44/ai/agents/execute/optional_packages.py
```diff
@@ -0,0 +1,222 @@
...
+    @property
+    def prompt_dir(self) -> str:
+        return self.template_dir or self.name
+
+    def prompt_template(self, prompt: OptionalPackagePrompt) -> str:
+        return f"optional_packages/{self.prompt_dir}/{prompt.value}.jinja2"
+
+
+OPTIONAL_PACKAGES: dict[str, OptionalPackage] = {
+    "mui": OptionalPackage(
```
morbargig: Copied from @88roy88 on PR #136: this is very hard to maintain like so, use the proposed new structure instead and import all of them.

Original: https://github.com/88roy88/FlowBolt/pull/136#discussion_r3412836216

backend/src/flow44/ai/agents/execute/templates/optional_packages/mui/codegen_rules.jinja2
```diff
@@ -0,0 +1,8 @@
+Required:
```
morbargig: Copied from @88roy88 on PR #136: can we change the structure a bit? maybe lets have an optional-pacakges folder that is not inside a specific agent, and it will have all of the different jinja, and maybe a python file with links to them.

Original: https://github.com/88roy88/FlowBolt/pull/136#discussion_r3412830392

backend/src/flow44/ai/generated_app_contract.py
```diff
@@ -0,0 +1,121 @@
...
+    "bun.lock",
```
morbargig: Copied from @88roy88 on PR #136: why are we protecting bun files??

Original: https://github.com/88roy88/FlowBolt/pull/136#discussion_r3412178493

backend/src/flow44/ai/generated_app_contract.py
```diff
@@ -0,0 +1,121 @@
...
+    "src/vite-env.d.ts",
+}
+_PROTECTED_SRC_DIRS = {
+    ("src", "api"),
+    ("src", "auth"),
+    ("src", "datasources"),
+    ("src", "platform"),
+}
+_PROTECTED_ROOT_DIRS = {
+    ".circleci",
```
morbargig: Copied from @88roy88 on PR #136: same?

Original: https://github.com/88roy88/FlowBolt/pull/136#discussion_r3412179622

backend/src/flow44/sandbox/pnpm_mixin.py
morbargig: we can backup the selected packages at the db to make it more clear.
but we still should work across the actual state off the project.
the `packge.json` is the one source of true.

and when  we have this logic we are prevent the ai builder to make wrong changes into those files.

morbargig: we can backup the packages but it will give us another source of true.

the backup on the db can be great for us to save some time when we want to inject the ai the parameters what are the packages that are install right now.

than the ai can chose if he need to explore more optimal packages or the current are satisfy

backend/src/flow44/sandbox/pnpm_mixin.py
morbargig: done

backend/src/flow44/ai/generated_app_contract.py
morbargig: done 
I removed them

backend/src/flow44/ai/generated_app_contract.py
morbargig: done
I removed them

backend/src/flow44/ai/agents/execute/templates/optional_packages/mui/codegen_rules.jinja2
morbargig: @88roy88 

do you want to push the OPTIONAL_PACKAGES object to a new file and link by object the jinja files?

backend/src/flow44/ai/generated_app_contract.py
morbargig: at first a did made a file marker inside the file  on an old version of this pr.
and than you asked me to remove the comments.
so i remove and change the logic.

so far this is the best result and i tested few.

backend/src/flow44/ai/agents/execute/optional_packages.py
morbargig: i think it's the best i can come up with if i want to keep this pr without more chnages.
i can reorder it lates on you can trust me.

if you insist to do it on some kind of logic please let me know.

morbargig (commented):
answer all comments

backend/src/flow44/ai/generated_app_contract.py
88roy88: copy from our convo:
אני חושב שיש ערך בלכתוב בקובץ עצמו שהוא לקריאה בלבד, כי תזכור שהbuilder שלנו קורא את הקבצים האלה. אני פשוט עוד לא בטוח מה הדרך הנכונה לוודא את מי מותר ואת מי אסור לערוך, אני מפחד שהגישה של תיקיה היא לא סקלבילית מצד אחד ושהשיטה של הלכתוב בתחילת הקובץ זה שביר מידי.

backend/src/flow44/ai/agents/execute/optional_packages.py
88roy88: this will make more sense after the work on the comment below about the restructure

backend/src/flow44/ai/agents/execute/templates/optional_packages/mui/codegen_rules.jinja2
88roy88: I want a folder agents/optional_packages 
and in it to have for example agents/optional_packages/leaflet
and in it to have a `__init__.py` file with the stuff from `"leaflet": OptionalPackage(` 
and all the jinja s files. and maybe in the __init__.py function to have also a class (with base class) for get_error_fix_prompt(**) that will return the relevant jinja template

backend/src/flow44/sandbox/pnpm_mixin.py
88roy88: do we allow the agent to add new packages during the run? or only at the planning level?

backend/src/flow44/ai/generated_app_contract.py
88roy88: I want the tool to block access to files and not just hope that the llm wont touch them. the question is how we identify files we dont want the llm to touch and if we even want it, if the llm decide there's something to add the the fetch.ts I dont think we should block it, but for the auth.ts maybe yes. we need to think which files to block and whats the best way to manage this list of files

morbargig: i took the blocked file from the prompt auth folder, vite.config, main.tsx.

those are files that the ai builder will like to make big changes with configuration and can cause others bugs.
as i mention i test this approach many times against few others like file marker inside the fle.

we can edit this files list as long as the project continue.
i didn't change to much and mostly i just force the old prompt version.

as we mention. to split prs.
this comment can be another task.
i was most to make sure the ai builder can't for 100% change files that are not for edit so this was the option that worked the best.

file marker inside the file will be hard and mussy to maintain.

morbargig: can we keep it this way for now, i know that to change this method will make the pr bigger and with less success precent.

backend/src/flow44/sandbox/pnpm_mixin.py
morbargig: for now the ai builder will only chose to add more packages only if in the first prompt.
not in the followup but we need to add it,
because iff the app grow and than the user release the app need to contains maps or router etc.

backend/src/flow44/ai/agents/execute/templates/optional_packages/mui/codegen_rules.jinja2
morbargig: it's nice approach, but is it mandatory for the pr? as i mention i want to make this pr easy as mush as possible.
i also had some other solution in my mind which i implemented earlier and i undo them to make the pr more ligth.

please answer me in the pr what is most for the pr, and what can be in phase 2, and what we still discuss about.

morbargig (commented):
self review

backend/src/flow44/ai/agents/execute/optional_packages.py
morbargig: please recheck

morbargig (commented):
func renaming

backend/src/flow44/ai/agents/execute/templates/optional_packages/mui/codegen_rules.jinja2
88roy88: yes. this is important for this pr.

backend/src/flow44/ai/generated_app_file_rules.py
```diff
@@ -0,0 +1,30 @@
...
+    "src/main.tsx",
+    "src/vite-env.d.ts",
+}
+PROTECTED_SRC_DIRS = {
+    ("src", "api"),
+    ("src", "auth"),
+    ("src", "datasources"),
+    ("src", "platform"),
+}
+PROTECTED_ROOT_DIRS = {
```
88roy88: I dont think the app have access to any of those...

backend/src/flow44/ai/agents/plan/templates/architecture.jinja2
```diff
@@ -26,6 +26,8 @@ Rules:
 - All file paths are relative to the project root (e.g. src/App.tsx)
 - Include only the files that need to be created or modified
 - **CRITICAL**: ONLY use the pre-installed packages (react, react-dom, @types/react, @types/react-dom, typescript, vite, tailwindcss, autoprefixer, postcss). Do NOT suggest adding other packages or dependencies. All functionality must be implemented using built-in browser APIs and these pre-configured packages only.
+
+{% include "_file_safety_rules.jinja2" %}
```
88roy88, line 30: where this resolves to? there is no such file here

backend/src/flow44/ai/agents/plan/prompts.py
```diff
@@ -4,15 +4,21 @@
 from pathlib import Path
 from typing import Any
 
-from jinja2 import Environment, FileSystemLoader
+from jinja2 import ChoiceLoader, Environment, FileSystemLoader
+
+from flow44.ai.generated_app_contract import generated_app_path_safety_prompt_context
 
 _templates_dir = Path(__file__).parent / "templates"
+_execute_templates_dir = Path(__file__).parents[1] / "execute" / "templates"
```
88roy88: so you resolve this to the execute agent, this is very ugly solution bro

backend/src/flow44/ai/agents/plan/prompts.py
```diff
@@ -4,15 +4,21 @@
...
-    loader=FileSystemLoader(str(_templates_dir)), trim_blocks=True, lstrip_blocks=True
+    loader=ChoiceLoader([FileSystemLoader(str(_templates_dir)), FileSystemLoader(str(_execute_templates_dir))]),
+    trim_blocks=True,
+    lstrip_blocks=True,
 )
 
 
 def render(template_name: str, **kwargs: Any) -> str:
+    kwargs.setdefault("file_safety", generated_app_path_safety_prompt_context())
     return _env.get_template(template_name).render(**kwargs)
```
88roy88, line 55: if you want this, why are you just dumping it into kwargs

backend/src/flow44/ai/agents/followup/templates/followup.jinja2
```diff
@@ -23,4 +23,6 @@ Follow the EXPLORE → REASON → ACT pattern:
 - Prefer `edit_file` for small, targeted changes. Use `write_file` for new files or when most of the file changes.
 - Keep changes minimal and focused on what the user asked for.
 - If the user asks a question, answer it in text. Only make edits if the user wants changes.
-- Do NOT include shell commands.
\ No newline at end of file
+- Do NOT include shell commands.
+
+{% include "_file_safety_rules.jinja2" %}
```
88roy88, line 59: same

backend/src/flow44/ai/agents/followup/prompts.py
```diff
@@ -3,15 +3,21 @@
 from pathlib import Path
 from typing import Any
 
-from jinja2 import Environment, FileSystemLoader
+from jinja2 import ChoiceLoader, Environment, FileSystemLoader
+
+from flow44.ai.generated_app_contract import generated_app_path_safety_prompt_context
 
 _templates_dir = Path(__file__).parent / "templates"
+_execute_templates_dir = Path(__file__).parents[1] / "execute" / "templates"
```
88roy88: same

backend/src/flow44/ai/agents/followup/prompts.py
```diff
@@ -3,15 +3,21 @@
...
 _env = Environment(  # noqa: S701 — templates are LLM prompts, not HTML; autoescape would break them
-    loader=FileSystemLoader(str(_templates_dir)), trim_blocks=True, lstrip_blocks=True
+    loader=ChoiceLoader([FileSystemLoader(str(_templates_dir)), FileSystemLoader(str(_execute_templates_dir))]),
+    trim_blocks=True,
+    lstrip_blocks=True,
 )
 
 
 def render(template_name: str, **kwargs: Any) -> str:
+    kwargs.setdefault("file_safety", generated_app_path_safety_prompt_context())
```
88roy88: same

backend/src/flow44/ai/generated_app_file_rules.py
```diff
@@ -0,0 +1,30 @@
...
+    ".github",
+    ".gitlab",
+    "backend",
+}
+PROTECTED_NAME_PREFIXES = (".env", "vite.config.")
+PROTECTED_NAME_SUBSTRINGS = ("template-guard", "template_guard")
```
88roy88: I dont understand why you dont also expose one ALL_PROTECTED or something so you dont have to import all and check each....

backend/src/flow44/ai/generated_app_file_rules.py
```diff
@@ -0,0 +1,30 @@
...
+    ("src", "auth"),
+    ("src", "datasources"),
+    ("src", "platform"),
+}
+PROTECTED_ROOT_DIRS = {
+    ".github",
+    ".gitlab",
+    "backend",
+}
+PROTECTED_NAME_PREFIXES = (".env", "vite.config.")
```
88roy88: why some are sets and some are tuples

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -111,6 +112,10 @@ async def write_file(path: str, content: str) -> str:
             """Write the full content of a file, creating it if needed. For small changes, prefer edit_file."""
             import difflib  # noqa: PLC0415
```
88roy88: I think yoni changed it, are you rebased?

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -111,6 +112,10 @@ async def write_file(path: str, content: str) -> str:
...
 
+            try:
+                path = validate_generated_app_edit_path(path)
+            except GeneratedAppContractError as exc:
+                return f"Error: {exc}"
```
88roy88: why generic error msg?

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -141,6 +146,10 @@ async def edit_file(path: str, search: str, replace: str) -> str:
             except FileNotFoundError:
                 return f"Error: File not found: {path}"
 
+            try:
+                path = validate_generated_app_edit_path(path)
+            except GeneratedAppContractError as exc:
+                return f"Error: {exc}"
```
88roy88: same

backend/src/flow44/ai/agents/execute/optional_packages.py
88roy88: not good

88roy88: we want the template not to be in one specific agent folder

backend/src/flow44/ai/agents/execute/templates/optional_packages/mui/codegen_rules.jinja2
morbargig: got it thanks

backend/src/flow44/ai/agents/execute/optional_packages.py
morbargig: @88roy88 
those changes are in another pr now...

backend/src/flow44/ai/agents/execute/templates/optional_packages/mui/codegen_rules.jinja2
morbargig: this is also not relevant to this pr any more...

backend/src/flow44/ai/generated_app_file_rules.py
morbargig: Ok you are right inside the workspace those folder are not exists.

backend/src/flow44/ai/agents/plan/templates/architecture.jinja2
morbargig: image

please check this path

morbargig: https://github.com/88roy88/FlowBolt/blob/a174878fd15c3acae3a9c167487760d4631e7585/backend/src/flow44/ai/agents/execute/templates/_file_safety_rules.jinja2

backend/src/flow44/ai/agents/plan/prompts.py
morbargig: I moved it to another folder

morbargig: Resolved in the top stack PR (#142) with commit 0b96afb. Shared prompt partials and optional-package prompt fragments now live under `flow44.ai.agents` shared ownership, `_execute_templates_dir` is removed, and non-execute agents no longer import from execute-owned prompt/package modules.

backend/src/flow44/ai/agents/followup/prompts.py
morbargig: those templates are shared so i just move them to another folder

morbargig:
Correction after splitting the stack cleanup: the execute-template path concern is now resolved directly in PR #138 with commit c4976fd. `_file_safety_rules.jinja2` was moved to `flow44.ai.agents.templates`, and plan/followup/fix_error/execute now load shared agent templates instead of resolving through `execute/templates`.

backend/src/flow44/ai/generated_app_file_rules.py
88roy88: so why dont you remove them?!

backend/src/flow44/ai/generated_app_file_rules.py
88roy88: what about that??

backend/src/flow44/ai/generated_app_file_rules.py
88roy88: so?

backend/src/flow44/ai/generated_app_contract.py
```diff
@@ -0,0 +1,68 @@
...
+        "protected_name_patterns": [f"{prefix}*" for prefix in PROTECTED_NAME_PREFIXES]
+        + [f"*{substring}*" for substring in PROTECTED_NAME_SUBSTRINGS],
+    }
+
+
+def normalize_generated_app_path(path: str) -> str:
+    """Normalize a generated path and reject absolute or traversal paths."""
+    if path.startswith(("/", "\\")):
+        raise GeneratedAppContractError(f"Unsafe generated file path: {path}")
+    normalized = path.replace("\\", "/").lstrip("/")
```
88roy88: is this for windows envs?

backend/src/flow44/ai/generated_app_contract.py
```diff
@@ -0,0 +1,68 @@
...
+        name in PROTECTED_FILES
+        or normalized.lower() in PROTECTED_SRC_PATHS
+        or name.startswith(PROTECTED_NAME_PREFIXES)
+        or parts[0] in PROTECTED_ROOT_DIRS
+        or (len(parts) >= 2 and parts[:2] in PROTECTED_SRC_DIRS)
+        or any(substring in name for substring in PROTECTED_NAME_SUBSTRINGS)
+    )
+
+
+def validate_generated_app_edit_path(path: str) -> str:
```
88roy88: rename function, to validate_generated_app_path_allowed or something in the same convention

backend/src/flow44/ai/agents/plan/prompts.py
```diff
@@ -4,15 +4,21 @@
 from pathlib import Path
 from typing import Any
 
-from jinja2 import Environment, FileSystemLoader
+from jinja2 import ChoiceLoader, Environment, FileSystemLoader
+
+from flow44.ai.generated_app_contract import generated_app_path_safety_prompt_context
 
 _templates_dir = Path(__file__).parent / "templates"
+_shared_templates_dir = Path(__file__).parents[1] / "templates"
```
88roy88: add a const you can import for the TEMPLATE_PROMPTS_PATH or something like that

88roy88 (changes requested):

backend/src/flow44/ai/agents/plan/prompts.py
morbargig: the raw is set-default,
if you overwrite it by the kwargs the default will not apply 

what do expect me to change ?

backend/src/flow44/ai/agents/execute/optional_packages.py
88roy88: right...

backend/src/flow44/ai/agents/execute/templates/optional_packages/mui/codegen_rules.jinja2
88roy88: ok, doesnt mean you dont have to fix it....

morbargig (commented):
fix all comments

backend/src/flow44/ai/agents/execute/templates/optional_packages/mui/codegen_rules.jinja2
morbargig: ok in the next pr

backend/src/flow44/ai/generated_app_file_rules.py
morbargig: i update it now

backend/src/flow44/ai/agents/plan/prompts.py
morbargig: ok i just convert it have type declaration for each render branch

backend/src/flow44/ai/agents/followup/templates/followup.jinja2
morbargig: can you explain ?

backend/src/flow44/ai/agents/followup/prompts.py
morbargig: recheck now tell me if you like it

backend/src/flow44/ai/generated_app_file_rules.py
morbargig: image

the ai create here some logic i just read it and i think it nice.

and this is why the constants are in another format.

backend/src/flow44/ai/generated_app_file_rules.py
morbargig: I warp it in some object.

backend/src/flow44/ai/agents/followup/agent.py
morbargig: yes i am rebased with develop

backend/src/flow44/ai/agents/followup/agent.py
morbargig: done

backend/src/flow44/ai/agents/followup/agent.py
morbargig: done

backend/src/flow44/ai/generated_app_contract.py
morbargig: yes it's for windows, do we need it yes?

backend/src/flow44/ai/generated_app_contract.py
morbargig: done

backend/src/flow44/ai/agents/plan/prompts.py
morbargig: done

morbargig (commented):
fix all comments

morbargig (commented):
.

backend/src/flow44/ai/generated_app_contract.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class GeneratedAppPathSafetyPromptContext(TypedDict):
+    protected_files: list[str]
+    protected_src_paths: list[str]
+    protected_src_dirs: list[str]
+    protected_name_patterns: list[str]
+
+
+def generated_app_path_safety_prompt_context() -> GeneratedAppPathSafetyPromptContext:
```
88roy88: tbh, not sure why this is not part of the `PROTECTED_APP_FILE_RULES` class

backend/src/flow44/ai/generated_app_contract.py
88roy88: and dont `pathlib.Path` or `os.path.normpath()` works?
in general I prefer to work with paths as Path and not str whenever possible

backend/src/flow44/ai/generated_app_contract.py
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+
+def normalize_generated_app_path(path: str) -> str:
+    """Normalize a generated path and reject absolute or traversal paths."""
+    if path.startswith(("/", "\\")):
+        raise GeneratedAppContractError(f"Unsafe generated file path: {path}")
+    normalized = path.replace("\\", "/").lstrip("/")
+    parts = PurePosixPath(normalized).parts
+    if not parts or any(part in {"", ".", ".."} for part in parts):
```
88roy88: this is against path traversal I guess but why isnt ./filename legit? (unlike ../filename)

backend/src/flow44/ai/generated_app_contract.py
```diff
@@ -0,0 +1,75 @@
...
+        raise GeneratedAppContractError(f"Unsafe generated file path: {path}")
```
88roy88: also, this mean that this function that claim to only normalize is also validating stuff which is weird

backend/src/flow44/ai/agents/followup/agent.py
88roy88: ok, can you do me a favor and move the import to the top of the file with the other imports? 
(I thought Yoni fixed it)

backend/src/flow44/ai/agents/execute/agent.py
```diff
@@ -265,16 +270,29 @@ async def _build_technical_plan(self, state: ExecutionState) -> WorkPlan:
...
+                    logger.warning("[execute] Dropping protected file from generated plan: %s", path)
+            if not safe_files:
+                continue
+            tasks.append(
+                Task(
+                    id=task_data.get("id", f"task-{uuid.uuid4().hex[:6]}"),
+                    title=task_data.get("title", "Untitled task"),
+                    description=task_data.get("description", ""),
+                    files=safe_files,
+                    depends_on=task_data.get("depends_on", []),
```
88roy88: and if it depends on a task you dropped?

88roy88 (changes requested):

backend/src/flow44/ai/agents/followup/agent.py
morbargig: done

backend/src/flow44/ai/generated_app_contract.py
morbargig: I research it and find that is the best PureWindowsPath


`Path` / `os.path.normpath()` work, but their behavior depends on the host OS. Here the input path may be generated with either Windows or POSIX separators, while the backend may run on Linux/WSL/macOS/Windows, so I wanted this contract to behave deterministically across hosts.

That’s why I used `PureWindowsPath` for parsing: it treats both `\` and `/` as separators and also lets us reject Windows drives / rooted / UNC paths explicitly. Then we canonicalize the accepted relative path with `PurePosixPath`, since the generated app contract stores and compares paths internally as forward-slash strings.

So this is still pathlib-based, just using pure path classes instead of host-dependent `Path`.

backend/src/flow44/ai/generated_app_contract.py
morbargig: ok done

backend/src/flow44/ai/generated_app_contract.py
morbargig: I added more tests and check it, i see it's cover all use cases

backend/src/flow44/ai/generated_app_contract.py
morbargig: ok i removed duplication and reused the code.

I split the safety parsing into _safe_generated_app_path_parts() and renamed the public helper to validate_and_normalize_generated_app_path() so callers explicitly opt into validation + canonicalization.

morbargig: done

backend/src/flow44/ai/agents/execute/agent.py
morbargig: this is a great comment, i fix it.

morbargig: done

morbargig (commented):
fix all last cr comments

88roy88 (approved):

backend/src/flow44/db/project_data_source.py
```diff
@@ -40,6 +41,15 @@ class DataSourceContext(DataSourceBase):
     module_path: str = ""
     generated_files: dict[str, str] = Field(default_factory=dict)
 
+    def to_prompt_context(self) -> dict[str, Any]:
```
88roy88, line 47: maybe im missing something, but whats this?

shalom-raichman: it just to remove some ugly and duplicate code from exe and plan agent, not new code new abstraction

backend/src/flow44/ai/generated_app_file_rules.py
```diff
@@ -0,0 +1,74 @@
...
+
+class ProtectedAppFileRules(NamedTuple):
+    files: frozenset[str]
+    src_paths: frozenset[str]
+    src_dirs: frozenset[tuple[str, ...]]
+    name_prefixes: frozenset[str]
+    name_substrings: frozenset[str]
+
+    def is_protected(self, normalized_path: str) -> bool:
+        """Return whether AI generation may not edit a normalized POSIX path."""
```
88roy88: @shalom-raichman can you do me a favor, do a quick pass and remove all ai comments and docs from all functions in this pr :)

backend/src/flow44/ai/generated_app_contract.py
```diff
@@ -0,0 +1,56 @@
+"""Backend-enforced file safety contract for AI-generated app files."""
+
+from __future__ import annotations
+
```
88roy88: also remove the file docs and `from __future__ import annotations` :)

backend/src/flow44/ai/generated_app_file_rules.py
shalom-raichman: done

backend/src/flow44/ai/generated_app_contract.py
shalom-raichman: done,
in this python version we don't need it anywhere although it used everywhere

88roy88: AI really like adding it for no reason

backend/src/flow44/ai/file_safety.py
```diff
@@ -0,0 +1,68 @@
...
+    path = PurePosixPath(normalized_path)
+    return any(path.full_match(p, case_sensitive=False) for p in _PROTECTED_APP_FILE_PATTERNS)
+
+
+def protected_file_rules() -> ProtectedFileRules:
+    paths: list[str] = []
+    patterns: list[str] = []
+    for pattern in _PROTECTED_APP_FILE_PATTERNS:
+        display = pattern.removeprefix("**/").replace("/**", "/*")
+        (patterns if "*" in display else paths).append(display)
```
88roy88: wow this is ugly, this is not JavaScript here, we are not doing this "awesome" one-liners here :)

backend/src/flow44/ai/file_safety.py
```diff
@@ -0,0 +1,83 @@
...
+    "**/*template_guard*",
+)
+
+
+def _is_protected(normalized_path: str) -> bool:
+    path = PurePosixPath(normalized_path)
+    return any(path.full_match(p, case_sensitive=False) for p in _PROTECTED_APP_FILE_PATTERNS)
+
+
+def protected_file_rules() -> ProtectedFileRules:
```
88roy88, line 52: in general feels to me like you over did it here, you could just have 
_PROTECTED_APP_FILE_PATTERNS
and 
_PROTECTED_APP_FILE_PATHS

backend/src/flow44/ai/file_safety.py
```diff
@@ -0,0 +1,83 @@
...
+    normalized = parsed.as_posix()
+    if _is_protected(normalized):
+        raise FileSafetyError(f"AI generation may not edit protected file: {normalized}")
+    return normalized
+
+
+def screen_generated_files(
+    generated: list[tuple[str, str]],
+    *,
+    source: str,
```
88roy88: not a fan of this `source` param

88roy88 (changes requested):

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -26,6 +27,13 @@
 MAX_READ_LINES = 1000
 
 
+def _format_file_safety_error(exc: FileSafetyError) -> str:
+    return (
+        f"Generated app contract violation: {exc}\n"
+        "Pick an editable app source file and only import packages selected for this project."
```
88roy88: the part about `import packages selected for this project` isnt clear to me

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -160,23 +164,27 @@ async def edit_file(path: str, search: str, replace: str) -> str:
...
-            if diff_str:
-                self._diffs.append(FileDiff(path=path, diff=diff_str))
-            if path not in self._files_changed:
-                self._files_changed.append(path)
+            await self._record_file_change(path, current, new_content, is_new=False)
             return f"OK — edited {path}"
 
         return ToolExecutor([grep, glob, read_file, write_file, edit_file])
 
+    async def _record_file_change(self, path: str, old_content: str, new_content: str, *, is_new: bool) -> None:
```
88roy88: nice 🍒

backend/src/flow44/ai/agents/execute/agent.py
```diff
@@ -261,16 +277,36 @@ async def _build_technical_plan(self, state: ExecutionState) -> WorkPlan:
...
         plan_data = parse_json_response(raw)
 
-        tasks = [
-            Task(
-                id=t.get("id", f"task-{uuid.uuid4().hex[:6]}"),
-                title=t.get("title", "Untitled task"),
-                description=t.get("description", ""),
-                files=t.get("files", []),
-                depends_on=t.get("depends_on", []),
+        tasks: list[Task] = []
```
88roy88: with the `rejected_file_notes` and stuff, do we still need this logic? I dont believe we want to just drop tasks, it will probably fail the build

backend/src/flow44/ai/agents/execute/execution_state.py
```diff
@@ -26,6 +26,7 @@ class ExecutionState(BaseModel):
     build_errors: str = ""
     all_errors: str = ""
     fix_attempts: int = 0
+    rejected_file_notes: list[str] = Field(default_factory=list)
```
88roy88: not sure I understand why to save it in the state, but I might be missing something

88roy88: (specifically in this agent where you dont have a step with validate_step like in the fix_error agent for example)

backend/src/flow44/ai/agents/fix_error/agent.py
```diff
@@ -219,25 +228,34 @@ async def _step_retry(self, state: FixErrorState) -> FixErrorState:
         )
 
         prompt = render_fix_errors(errors=state.validation_errors, files=dict(state.generated_files))
+        messages: list[dict[str, Any] | Message] = [Message.user("Fix the TypeScript errors.")]
+        if state.rejected_file_notes:
```
88roy88: not sure I would put it in the same step, having a message of 
- x is borken
- something something not working.
- "Fix the TypeScript errors."
- fixing something something but editing a protected file.
- "Fix the TypeScript errors." (not great...)
- .... 

^^ the last `"Fix the TypeScript errors."` wasnt really a TypeScript error I think..

backend/src/flow44/ai/file_safety.py
```diff
@@ -0,0 +1,83 @@
...
+) -> tuple[list[tuple[str, str]], list[str]]:
```
88roy88: this is a very ugly return type. 
depending on what you are going to do in the executor we might not even need it.
and converting the exceptions to str is also not a best practice

backend/src/flow44/ai/file_safety.py
```diff
@@ -0,0 +1,83 @@
...
+    paths: list[str] = []
+    patterns: list[str] = []
+    for pattern in _PROTECTED_APP_FILE_PATTERNS:
+        display = pattern.removeprefix("**/").replace("/**", "/*")
+        (patterns if "*" in display else paths).append(display)
+    return {"protected_paths": sorted(paths), "protected_patterns": sorted(patterns)}
+
+
+def normalized_path_or_reject(path: str) -> str:
+    parsed = PureWindowsPath(path)
```
88roy88, line 60: why do we need the windows stuff?

88roy88 (changes requested):

backend/src/flow44/ai/agents/execute/agent.py
shalom-raichman: should we replan it with a rejection note and replan ?

backend/src/flow44/ai/file_safety.py
shalom-raichman: done

backend/src/flow44/ai/file_safety.py
shalom-raichman: done

backend/src/flow44/ai/file_safety.py
shalom-raichman: it's was originally for the error message but we can do without it

done

backend/src/flow44/ai/agents/followup/agent.py
shalom-raichman: probably left over from the spaghetti prs, I have rephrased it

done

backend/src/flow44/ai/agents/followup/agent.py
shalom-raichman: thank you :)

backend/src/flow44/ai/agents/execute/agent.py
shalom-raichman: I have removed this logic, and added validation to the rejection files too on top of the build clean vlidation, and a user message if the retry ended with the rejected_files not empty, so we mostly counting on the write time rejection but don't leave any rejected files unreported back to the agent.

backend/src/flow44/ai/agents/execute/execution_state.py
shalom-raichman: we do have validate state on line 61 in this file, and with the previous comment on the task logic been solved we now rely even more on the rejected file state

backend/src/flow44/ai/file_safety.py
shalom-raichman: done

backend/src/flow44/ai/file_safety.py
shalom-raichman: if the agent got confused and return windows path, just defense in depth

backend/src/flow44/ai/agents/fix_error/agent.py
shalom-raichman: I have change the terminology of the fix_errors to feedback, and now he have to feedback sections

88roy88:
This PR now conflicts with `develop` (currently `mergeable_state: dirty`) because of an overlap with #157, which merged after this branch was created.

**Root cause:** both PRs independently refactored the same inline diff-tracking code in `followup/agent.py` and `fix_error/agent.py`:

- This PR (#138) extracted the old inline `difflib.unified_diff` + `self._diffs: list[FileDiff]` logic into a per-agent `_record_file_change()` helper, and added `normalized_path_or_reject()` file-safety checks in `write_file`/`edit_file`.
- #157 replaced the same inline logic with a shared abstraction on a new `ChatAgent` base class (`_write_file_and_emit_diff()` / `_edit_file_and_emit_diff()`, writing into a new `DiffTracker` that also collapses repeat writes to the same file into one diff). It also changed both agents' base class from `BaseAgent` to `ChatAgent` and deleted the old `FileDiff` class from `followup/agent.py`.

Since #157 already merged, this branch's `_record_file_change`/`FileDiff`/`BaseAgent` world no longer exists on `develop` — resolving this needs more than picking one side per conflicting hunk.

**To resolve:**
1. Merge/rebase from `develop` to pick up `ChatAgent` and `DiffTracker`.
2. Re-thread the `normalized_path_or_reject()` file-safety check into `_write_file_and_emit_diff()` / `_edit_file_and_emit_diff()` (or call it before invoking them) instead of the now-deleted `_record_file_change`.
3. In `fix_error/agent.py`'s retry loop, reconcile this PR's `screen_generated_files`/`validated` file-safety screening with #157's `_write_file_and_emit_diff` call over `generated` — right now they're two different loops over the generated files that both need to survive.

Happy to pair on this if useful — flagging now since `develop`'s branch protection is `strict: true`, so this'll need to be resolved and green before it can merge regardless.

backend/src/flow44/ai/agents/execute/agent.py
```diff
@@ -261,17 +281,28 @@ async def _build_technical_plan(self, state: ExecutionState) -> WorkPlan:
...
+                title=task_data.get("title", "Untitled task"),
+                description=task_data.get("description", ""),
+                files=task_data.get("files", []),
+                depends_on=task_data.get("depends_on", []),
             )
-            for t in plan_data.get("tasks", [])
+            for task_data in plan_data.get("tasks", [])
         ]
 
+        task_ids = {task.id for task in tasks}
```
88roy88, line 296: just this part feels a bit iffy to me, dont we want to handle it better by telling the agents to revise the work plan instead of dropping maybe importnat files?

88roy88 (changes requested):

backend/src/flow44/ai/agents/execute/agent.py
88roy88: decided to add a comment here that we dont have guard here and will also open an issue for it, waiting for it to happen in prod :)

88roy88 (approved):

PR #149, opened 2026-06-23
feat: git-backed version control, guarded by a workspace lock

shalom-raichman:
## What this adds

Git-backed version control for project workspaces. Every AI turn commits a version; users can preview an older version, restore to it, and save or discard their own edits as versions of their own. The chat sidebar stamps each assistant message with the version it produced.

Versioning lives in `services/versioning/` — a thin `Git` wrapper (`git.py`) over `--git-dir`/`--work-tree` subprocesses, and a `service.py` that owns the orchestration and the DB events clients render from.

## The workspace lock

The first half of this branch shipped versioning by accretion: each pitfall got its own guard, and the guards ended up disagreeing. The symptoms all traced to one missing idea — nothing said **who owns the working tree right now**.

- `commit_turn` ended with `add -A`, so any user edit left uncommitted when Send was pressed was swept into the AI's version and attributed `author="ai"`.
- The five mutating routes in `api/files.py` were gated by static RBAC only. The UI blocked; the server did not.
- New and legacy projects initialised git down different paths, with an `is_repo()` branch in every public function.
- Dirtiness was resolved as a side effect of *navigating versions* (`on_dirty="save"|"discard"` threaded through preview and restore), so discarding work was something that happened to the user rather than something the user did.

The invariant now installed:

> The working tree is dirty only while a human deliberately made it so, and every operation that needs a clean tree refuses rather than resolving it on the user's behalf.

It stores **nothing new**. It is a pure function of three facts already available: `is_run_active(project_id)`, `git.is_detached()`, `git.is_dirty()`.

### What that changed

| Area | Change |
|---|---|
| `git.py` | `_head()` reads `.git/HEAD` once; `is_repo`/`is_detached` become sync file reads instead of 1–2 subprocesses each. `_succeeds` deleted. `clean_untracked()` → `discard_all()` (`reset --hard` + `clean -fd`). |
| `service.py` | One `WorkspaceLocked(code, files)` replaces `RunActiveError` / `PreviewActiveError` / `DirtyWorkspaceError`. One `_workspace(project_id, needs)` door does `ensure_repo` + precondition + failure-resync. `handle_version_op`, `_commit_unsaved_edits`, `_bootstrap_legacy`, `init_scaffold_version`, `_version_exists`, `_newest_version` and `on_dirty` all deleted, replaced by explicit named ops. |
| `deps.py` / `files.py` | `require_writable_workspace` returns 409 with the refusal payload on all five mutating routes. The guard is now server-side, not just UI. |
| `chat.py` | Five `elif` branches and three `except` clauses replace the per-error handler functions; `_start_turn` restores develop's flat turn-start indentation. |
| Frontend | `on_dirty` removed from the protocol; `discard_edits` added. `resolveDirtyOp` settles the tree, then re-issues the pending op on the same socket. The unsaved-edits dialog lists the files Discard would destroy. A 409 from a file write surfaces through `useErrorStore`. |

### Preconditions

Three states, three preconditions — composed from two primitives, so no condition is asked twice:

- `_needs_writable` (idle + attached) — file writes, `save_version`, `discard_edits`
- `_needs_settled` (idle + clean) — `preview_version`, `restore_version`; both are things you do *from* a preview, so they must not require an attached HEAD
- `_needs_turn` (both) — `begin_turn`

### Behaviour changes

- Pressing Send with unsaved edits is **refused**, with a dialog listing the changed files. Save commits them as an `author="user"` version first, so the AI's version then contains only the AI's changes. Discard drops them. Nothing is resolved on the user's behalf.
- Writing a file while an agent runs, or while previewing an old version, returns **409** rather than silently succeeding.
- Connecting to a project that predates versioning now creates its repo and announces "Version 0". That is how old projects get versioning.

## Known gap

`api/terminal.py` is a full interactive PTY gated only by role. It is the one channel that can still dirty the tree unobserved. Operations refuse over a tree it dirtied rather than corrupting it, but nothing prevents it. Out of scope here.

Human-vs-human co-editing also stays unprevented — Monaco buffers plus whole-file PUT are already last-write-wins. That needs per-file presence, not a workspace mutex.

## ⚠️ Post-merge step

**After merging, run `POST /api/admin/maintenance/sync-protected-files` (admin only).**

`ensure_repo` runs `git add -A` on legacy workspaces, and an old workspace may have no `.gitignore` at all. `**/.gitignore` is in `SYNCABLE_TEMPLATE_FILE_PATTERNS` (`ai/file_safety.py`), so this endpoint distributes the template's `.gitignore` — which ignores `node_modules` — to every existing workspace. `_seed_gitignore` is the in-code safety net for fresh inits so correctness never depends on someone remembering; this run is the fleet-wide fix for workspaces that already have a repo.

## Testing

```bash
cd backend
uv run --no-sync pytest tests/versioning tests/api tests/sandbox tests/db --no-cov
uv run --no-sync task lint

pnpm -C frontend test && pnpm -C frontend typecheck && pnpm -C frontend lint
```

Regressions with a test that pins them, none of which may be recreated:

1. A workspace without `.git` reaching the enclosing repo (`--git-dir`/`--work-tree`; `_head()` is per-workspace by construction).
2. `head_sha()` leaking the outer repo's HEAD, giving legacy projects a false preview banner.
3. A half-built `.git/` holding only `objects/` is not a repo, and `init` must adopt it.
4. The selector event loop has no subprocess support → `asyncio.to_thread`.
5. Restore leaving `main` ahead → the single `checkout -f -B main <sha>`.
6. preview → restore → preview → exit stays on the restored version.
7. A refused turn leaves no chat row, no event, and no deleted pending plan.
8. Restore keeps its events when git fails.
9. `commit_turn` refuses while detached.
10. A user edit saved before a refused turn produces its own `author="user"` version, and the AI's next version contains only the AI's changes.


backend/src/flow44/api/chat.py
```diff
@@ -110,16 +113,42 @@ async def _run_agent_safe(project_id: str, coro: Any, claimed_at: datetime) -> N
...
     if claimed_at is None:
         coro.close()
         raise AgentAlreadyRunning
-    task = asyncio.create_task(_run_agent_safe(project_id, coro, claimed_at))
+    task = asyncio.create_task(_run_agent_safe(project_id, sandbox, coro, claimed_at))
     _RUNNING.add(task)
     task.add_done_callback(_RUNNING.discard)
 
 
+async def _reject_edit_while_previewing(websocket: WebSocket, sandbox: PnpmSandbox, project_id: str) -> bool:
```
shalom-raichman: with the new service convention in our repo we should move these functions to a service file in the services folder

backend/src/flow44/api/chat.py
```diff
@@ -110,16 +113,42 @@ async def _run_agent_safe(project_id: str, coro: Any, claimed_at: datetime) -> N
...
+    """Refuse a build started while previewing an old version (detached HEAD)."""
```
shalom-raichman: remove docstring

backend/src/flow44/api/chat.py
```diff
@@ -110,16 +113,42 @@ async def _run_agent_safe(project_id: str, coro: Any, claimed_at: datetime) -> N
...
+) -> None:
+    if await is_run_active(project_id):
+        await websocket.send_json({"type": "error", "message": "Can't change versions while the AI is working"})
+        return
+    try:
+        if msg_type == "exit_preview":
+            await versioning.exit_preview(sandbox, project_id)
+        elif msg_type == "preview_version":
+            await versioning.preview_version(sandbox, project_id, data["commit_sha"])
+        else:  # restore_version
```
shalom-raichman: next line literally say what the comment explain, remove the comment

backend/src/flow44/db/events.py
```diff
@@ -66,18 +66,27 @@ class AgentEvent(SQLModel, table=True):
...
+        )
         session.add(row)
         await session.commit()
 
     if notify:
         await _notify(project_id, event)
 
 
+async def emit_transient(project_id: str, event: dict[str, Any]) -> None:
+    # Broadcast only — never persisted, so not replayed on history reload.
```
shalom-raichman: make this comment clearer but not longer

backend/src/flow44/db/events.py
```diff
@@ -90,7 +99,25 @@ async def get_events(project_id: str, after_id: int = 0) -> list[AgentEvent]:
...
     async with database.async_session() as session:
-        result = await session.execute(select(AgentEvent).where(AgentEvent.project_id == project_id))
-        for row in result.scalars().all():
-            await session.delete(row)
+        await session.execute(delete(AgentEvent).where(col(AgentEvent.project_id) == project_id))
+        await session.commit()
+
+
+async def get_versions(project_id: str) -> list[AgentEvent]:
+    """Return the project's version_committed events (the app version history)."""
```
shalom-raichman: remove docstring

backend/src/flow44/versioning/git_service.py
```diff
@@ -0,0 +1,107 @@
+"""Thin git wrapper for per-workspace app versioning.
+
```
shalom-raichman: delete docstring

backend/src/flow44/versioning/git_service.py
```diff
@@ -0,0 +1,107 @@
...
+import re
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from flow44.sandbox.main import PnpmSandbox
+
+logger = logging.getLogger(__name__)
+
+
+_GIT_NAME = "FlowBolt AI"
```
shalom-raichman: change to BuildApp also in the email and msg

backend/src/flow44/versioning/git_service.py
```diff
@@ -0,0 +1,107 @@
...
+_GIT_EMAIL = "ai@flowbolt.local"
+
+# git message passed via file (commit -F), not -m: cmd.exe on Windows mangles quoted args.
```
shalom-raichman: check again the -f vs -m, windows do support git commit -m "some commit"

backend/src/flow44/api/chat.py
```diff
@@ -110,16 +113,59 @@ async def _run_agent_safe(project_id: str, coro: Any, claimed_at: datetime) -> N
         await clear_heartbeat(project_id, only_beat=beat.at)
 
 
-async def _start_agent(project_id: str, coro: Any) -> None:
-    claimed_at = await try_claim_run(project_id)
+async def _claim_run(project_id: str) -> datetime:
+    """Take the run lock up front, so a refused turn never reaches its side effects."""
```
shalom-raichman, line 120: remove docstring

backend/src/flow44/api/chat.py
```diff
@@ -110,16 +116,40 @@ async def _run_agent_safe(project_id: str, coro: Any, claimed_at: datetime) -> N
         await clear_heartbeat(project_id, only_beat=beat.at)
 
 
-async def _start_agent(project_id: str, coro: Any) -> None:
-    claimed_at = await try_claim_run(project_id)
+async def _claim_run(project_id: str) -> datetime:
```
shalom-raichman, line 119: I don't like the fact that we changing core logic for the versioning feature

backend/src/flow44/api/chat.py
```diff
@@ -264,40 +280,55 @@ async def _receive_actions() -> None:  # noqa: C901, PLR0912, PLR0915
...
+                            fix_agent.run,
                             error_message=error_message,
                             error_file=error_file,
                             error_line=error_line,
                             error_stack=error_stack,
                         ),
+                        prepare=partial(_save_user_turn, project.id, error_desc, fix_event),
                     )
+
+                elif msg_type == "preview_version":
```
shalom-raichman: must be a better way to handle the messages instead of this hisios elifs

backend/src/flow44/api/deps.py
```diff
@@ -163,6 +165,16 @@ async def _check(user_permissions: PermissionsDep) -> set[Permission]:
     return Depends(_check)
 
 
+async def require_writable_workspace(project: ProjectDep) -> AsyncIterator[None]:
```
shalom-raichman, line 168: dont like the name and the fact that it needs docstring to explain what it do

backend/src/flow44/api/projects.py
```diff
@@ -103,6 +104,7 @@ async def _create() -> None:
 
         try:
             await sandbox_manager.create_sandbox(project.id)
+            await versioning.ensure_repo(project.id)
```
shalom-raichman, line 107: ensure applies that is no op, witch is not true, the function creates a repo if not exists

backend/src/flow44/db/chat.py
```diff
@@ -43,3 +43,15 @@ async def get_messages(project_id: str) -> list[ChatMessage]:
             select(ChatMessage).where(ChatMessage.project_id == project_id).order_by(col(ChatMessage.created_at).asc())
         )
         return list(result.scalars().all())
+
+
+async def trim_messages_after(project_id: str, created_at: str) -> None:
+    """Delete messages created after the given timestamp (used when a restore forks history)."""
```
shalom-raichman: remove docstring

backend/src/flow44/db/events.py
```diff
@@ -66,18 +70,27 @@ class AgentEvent(SQLModel, table=True):
...
+            created_at=datetime.now(UTC),
+        )
         session.add(row)
         await session.commit()
 
     if notify:
         await _notify(project_id, event)
 
 
+async def emit_transient(project_id: str, event: dict[str, Any]) -> None:
```
shalom-raichman, line 84: why not use notify directly, it just an abstraction layer without any reason

backend/src/flow44/services/versioning/git.py
```diff
@@ -0,0 +1,94 @@
+from __future__ import annotations
+
+import asyncio
+import os
+import subprocess
+from pathlib import Path
+
+from flow44.sandbox.base import workspace_path
+
+_GIT_NAME = "BuildApp AI"
```
shalom-raichman, line 11: should we have it in the config file with a new class of git workspace settings? or its an over kill

where else do we have constants  like that in the top of the file regarding to the versioning feature

backend/src/flow44/services/versioning/git.py
```diff
@@ -0,0 +1,94 @@
...
+    pass
+
+
+class Git:
+    def __init__(self, project_id: str) -> None:
+        self.workspace_dir = workspace_path(project_id)
+        self.git_dir = os.path.join(self.workspace_dir, ".git")
+
+    async def _run(self, *args: str) -> str:
+        # Naming the repo is what contains git: under bare `-C`, a workspace without `.git` finds the enclosing one.
```
shalom-raichman: trim the comments to the bear minimum

backend/src/flow44/services/versioning/git.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # -- Queries --
+
+    def _head(self) -> str | None:
+        try:
+            return Path(self.git_dir, "HEAD").read_text()
+        except OSError:
+            return None
+
+    def is_repo(self) -> bool:
```
shalom-raichman, line 60: do we really want this abstraction layer ?

backend/src/flow44/services/versioning/service.py
```diff
@@ -0,0 +1,244 @@
...
+
+from flow44.config import settings
+from flow44.db.chat import trim_messages_after
+from flow44.db.events import AgentEvent, emit_event, emit_transient, get_versions, trim_events_after
+from flow44.db.heartbeat import is_run_active, try_claim_run
+from flow44.services.versioning.git import Git
+
+logger = logging.getLogger(__name__)
+
+_SCAFFOLD_MESSAGE = "Version 0 — blank scaffold"
```
shalom-raichman, line 20: same as the git creds, nees to think if is better to move these into gite settings

backend/src/flow44/services/versioning/service.py
```diff
@@ -0,0 +1,244 @@
...
+
+async def _needs_turn(project_id: str, git: Git) -> None:
+    await _needs_writable(project_id, git)
+    await _refuse_if_dirty(git)
+
+
+Precondition = Callable[[str, Git], Awaitable[None]]
+
+
+def _seed_gitignore(workspace_dir: str) -> None:
```
shalom-raichman: why do we need it? we have updated the gitignore template and going to run the update syncbabel script

backend/src/flow44/services/versioning/service.py
```diff
@@ -0,0 +1,244 @@
...
+        await _emit_committed(project_id, sha)
+
+
+@asynccontextmanager
+async def _workspace(project_id: str, needs: Precondition | None = None) -> AsyncIterator[Git]:
+    async with _locks[project_id]:
+        git = Git(project_id)
+        await _ensure_repo(project_id, git)
+        if needs:
+            await needs(project_id, git)
```
shalom-raichman: needs is a name that is not self explanatory

backend/src/flow44/services/versioning/service.py
```diff
@@ -0,0 +1,244 @@
...
+    if git.is_repo():
+        return
+    _seed_gitignore(git.workspace_dir)
+    sha = await git.init(_SCAFFOLD_MESSAGE)
+    if sha:
+        await _emit_committed(project_id, sha)
+
+
+@asynccontextmanager
+async def _workspace(project_id: str, needs: Precondition | None = None) -> AsyncIterator[Git]:
```
shalom-raichman: new name too

backend/src/flow44/services/versioning/service.py
```diff
@@ -0,0 +1,244 @@
...
+
+def _find(versions: list[AgentEvent], commit_sha: str) -> AgentEvent | None:
+    return next((v for v in versions if v.payload["commit_sha"] == commit_sha), None)
+
+
+def _newest_sha(versions: list[AgentEvent]) -> str | None:
+    return versions[-1].payload["commit_sha"] if versions else None
+
+
+# -- Telling clients --
```
shalom-raichman, line 105: why 3 functions ?

backend/src/flow44/services/versioning/service.py
```diff
@@ -0,0 +1,244 @@
...
+    async with _workspace(project_id, _needs_turn):
+        return await try_claim_run(project_id)
+
+
+async def save_version(project_id: str) -> None:
+    async with _workspace(project_id, _needs_writable) as git:
+        files = await git.changed_files()
+        sha = await git.commit_all(_USER_EDITS_MESSAGE)
+        if sha:
+            await _emit_committed(project_id, sha, author="user", files=files)
```
shalom-raichman: needs to check who is the author, and when user commit manual changes if it is saved to the context

frontend/src/components/chat/cards/UserEditCard.tsx
```diff
@@ -0,0 +1,20 @@
+import { useTranslation } from 'react-i18next';
+import { Pencil } from 'lucide-react';
+import { CardWrapper } from './CardWrapper';
+
+export function UserEditCard({ files }: { files: string[] }) {
```
shalom-raichman: why not use the existing card of edits and make him generic ?

frontend/src/components/ui/action-button.tsx
```diff
@@ -0,0 +1,10 @@
+import { Button, type ButtonProps } from './button';
+
+// A disabled button fires no mouse events, so the tooltip has to live on a wrapper.
+export function ActionButton({ title, ...props }: ButtonProps & { title: string }) {
```
shalom-raichman, line 4: why need a custom button ?

frontend/src/components/ui/confirm-dialog.tsx
```diff
@@ -0,0 +1,57 @@
+import type { ReactNode } from 'react';
+import { useTranslation } from 'react-i18next';
+import { Button, type ButtonProps } from './button';
+import { Dialog, DialogContent, DialogClose, DialogTitle } from './dialog';
+
+interface ConfirmDialogProps {
```
shalom-raichman, line 6: why so many props ???

frontend/src/components/version/UnsavedEditsDialog.tsx
```diff
@@ -0,0 +1,32 @@
+import { useTranslation } from 'react-i18next';
+import { Save } from 'lucide-react';
+import { useVersionStore } from '../../stores/version';
+import { ConfirmDialog } from '../ui/confirm-dialog';
+import { DirtyFileList } from './DirtyFileList';
+
+export function UnsavedEditsDialog() {
```
shalom-raichman, line 7: why is the wrapper, is this a good abstraction or a can avoid one

frontend/src/services/api.ts
```diff
@@ -23,6 +23,20 @@ async function fetchWithAuth(path: string, options: RequestInit = {}, allowRetry
   return res;
 }
 
+export class ApiError extends Error {
```
shalom-raichman, line 26: what this have to here anyway? I don't think it's related

backend/src/flow44/api/chat.py
shalom-raichman: done

backend/src/flow44/api/chat.py
shalom-raichman: up to yoni if it's ok with a follow-up refactor or refactor here

backend/src/flow44/api/chat.py
shalom-raichman: done

backend/src/flow44/db/chat.py
shalom-raichman: done

backend/src/flow44/services/versioning/git.py
shalom-raichman: never mind is good as is

backend/src/flow44/services/versioning/git.py
shalom-raichman: done

backend/src/flow44/services/versioning/service.py
shalom-raichman: removed

backend/src/flow44/services/versioning/service.py
shalom-raichman: done

backend/src/flow44/services/versioning/service.py
shalom-raichman: added to the context

backend/src/flow44/api/projects.py
yoni5353: in my opinion "ensure" is fine, but you can change if u want

backend/src/flow44/api/deps.py
yoni5353: you were not happy with the name as it is now? i can't think of a concise name, tbh. maybe `require_idle_and_on_main` which is more specific but it's still a mouthful

backend/src/flow44/services/versioning/service.py
```diff
@@ -0,0 +1,247 @@
...
+            await save_message(project_id, ChatRole.user, _edit_note([d.path for d in diffs]))
+            await _emit_committed(project_id, sha, author="user", diffs=diffs)
+        await _emit_dirty(project_id, git)
+
+
+async def discard_edits(project_id: str) -> None:
+    async with _locked_workspace(project_id, [_not_busy, _not_detached]) as git:
+        await git.discard_all()
+        await _emit_current_version(project_id, git)
+        await _emit_dirty(project_id, git)
```
yoni5353: the name `_emit_dirty` is a little confusing, because in most places this toggles the "dirtiness" off and not on.

maybe at least name it something like `_emit_dirty_state` or `_emit_dirty_files`?

backend/src/flow44/services/versioning/service.py
```diff
@@ -0,0 +1,247 @@
...
+            diffs = await git.commit_diffs(sha, max_chars=_MAX_DIFF_CHARS)
+            await save_message(project_id, ChatRole.user, _edit_note([d.path for d in diffs]))
+            await _emit_committed(project_id, sha, author="user", diffs=diffs)
+        await _emit_dirty(project_id, git)
+
+
+async def discard_edits(project_id: str) -> None:
+    async with _locked_workspace(project_id, [_not_busy, _not_detached]) as git:
+        await git.discard_all()
+        await _emit_current_version(project_id, git)
```
yoni5353: the ai tells me you need this line to trigger `refreshEditorFiles()` through `handleVersionPreviewActive()` in frontend to refetch the files, which was not really clear to me at all, since discard_edits doesn't update the current version at all.

i think we should either explain this line in a comment, or change it to a different event. (`_emit_current_version` is for sure weird since it has a db query to get_versions, but we already know that we are on main here)

backend/src/flow44/api/chat.py
```diff
@@ -151,7 +180,9 @@ async def chat_ws(  # noqa: C901, PLR0915
         await websocket.close()
         return
 
+    had_subscribers = has_subscribers(project.id)
     queue = subscribe(project.id)
+    await versioning.broadcast_current_version(project.id, reset_orphaned_preview=not had_subscribers)
```
yoni5353: this condition is a bit iffy, because there could be subscribers on a different pod, and has_subscribers only checks for subscribers in-memory.

i don't know if there is much to do about this here, except not resetting the branch. it's probably insignificant enough to leave as-is if you like the feature

frontend/src/stores/files.ts
```diff
@@ -31,6 +32,15 @@ interface FilesState {
 
 let _fileTreeRequestSerial = 0;
 
+// The UI already blocks these, so a 409 is a real race the user has to see.
```
yoni5353: i did not understand this comment. the UI blocks what?

frontend/src/stores/version.ts
```diff
@@ -0,0 +1,192 @@
...
+  return true;
+}
+
+function issue(attempt: Attempt & { undo?: () => void }, message: WSMessage) {
+  settle(false);
+  pending = attempt;
+  if (!send(message)) settle(false);
+}
+
+export function armTurn(run: () => void, undo: () => void) {
```
yoni5353: maybe let's turn the signature to an objet param `armTurn({run, undo}: {...}), because from outside i didn't understand their roles, i thought it was a single list of effects.

frontend/src/components/ui/confirm-dialog.tsx
yoni5353: imo this is fine

frontend/src/components/ui/action-button.tsx
yoni5353: i agree, i would have removed this

frontend/src/components/layout/AppShell.tsx
```diff
@@ -76,6 +78,7 @@ export function AppShell() {
   };
 
   // Shared state
+  const { code } = useWorkspaceLock();
```
yoni5353: maybe rename this usage too `lockCode` or `workspaceLock` just because `code` is really not clear in this huge component

yoni5353: same for ErrorToast.tsx

yoni5353: and PromptInput.tsx

frontend/src/stores/chat.ts
```diff
@@ -240,19 +275,15 @@ export const useChatStore = create<ChatState>((set, get) => ({
...
-      } else {
-        // Read-only: replay events through a temporary handler without WS
-        const tempHandler = createSendMessageHandler(set, get, () => {});
-        for (const evt of events) {
-          tempHandler(evt as WSMessage);
-        }
+      const replayHandler = handler ?? createSendMessageHandler(set, get, () => {});
+      for (const evt of events) {
+        replayHandler(evt as WSMessage);
+        handleVersionMessage(evt as WSMessage);
```
yoni5353: you created a dedicated handler for handling version messages, but it could just be implemented in the regular old replayHandler (created by createSendMessageHandler), so that u won't need to call two handlers here and in line 278 above `socket.onMessage(handleVersionMessage);`

if u want the version handling grouped together in `version.ts`, u can also call `handleVersionMessage` inside the regular handler

backend/src/flow44/api/deps.py
shalom-raichman: left as is

frontend/src/components/ui/action-button.tsx
shalom-raichman: deleted

frontend/src/components/ui/confirm-dialog.tsx
shalom-raichman: deleted

backend/src/flow44/services/versioning/service.py
shalom-raichman: Renamed to `_emit_dirty_files`.

frontend/src/stores/files.ts
shalom-raichman: Reworded: the editing controls are disabled while the workspace is locked, so a 409 here means a race the user needs to see.

frontend/src/stores/version.ts
shalom-raichman: Changed to `armTurn({ run, undo })`.

frontend/src/components/layout/AppShell.tsx
shalom-raichman: Renamed the hook's return field to `lock` in `workspaceLock.ts`, so AppShell, ErrorToast, PromptInput and the others all read `lock` now.

backend/src/flow44/services/versioning/service.py
shalom-raichman: You're right, it was only there to refresh files. Discard now sends its own workspace_reset event. I also replaced the DB query with not git.is_detached(), the same check the lock uses.

frontend/src/stores/chat.ts
shalom-raichman: The send handler is per-turn: it's detached on action_complete/error, and fix-error turns use a different handler. Most version events (user-edit commits, preview, restore, workspace_dirty) arrive between turns, so they need a listener that stays attached for the whole session. If the send handler also called handleVersionMessage, every live version event would be processed twice during a turn (duplicate edit cards). The only clean way to merge them is a single long-lived router that sends each message to both handlers, which rewires attach/detach for every turn. I don't think that's worth it here.

btw I can reply to you with my own words if you like, I just think claude is better explaining things then me, let me know :)

backend/src/flow44/api/chat.py
shalom-raichman: You're right, has_subscribers only sees this pod's connections. I removed the reset (even though I really liked it): an orphaned preview now just shows the banner with "exit preview". The workspace lock had the same problem, so it now uses a Postgres advisory lock.

frontend/src/stores/chat.ts
yoni5353: claude is fine, as long as it's not too wordy :p

ok so from what i understand `replayHandler` listens for events after each message until it completes, and `handleVersionMessage` always listens.

that makes sense, but it wasn't intuitive for me while reading the code

claude suggested adding a comment like
```ts
// Version handler stays attached for the whole session; the agent handler detaches after each turn.
socket.onMessage(handleVersionMessage);
```

but i think we can make this clearer with renames, like changing `handleVersionMessage` to `handleOutOfTurnMessage`

or `replayHandler` -> `turnHandler` and `handleVersionMessage` -> `backgroundHandler`

yoni5353 (commented):
nice, i liked the changes.

only left one thread open about the message handlers

PR #154, opened 2026-06-28
Shalom/fix stalled agent

shalom-raichman:


backend/src/flow44/services/heartbeat_reaper.py
```diff
@@ -0,0 +1,49 @@
...
+import logging
+
+from flow44.config import settings
+from flow44.db.heartbeat import reap_stale_runs
+
+logger = logging.getLogger(__name__)
+
+
+class HeartbeatReaper:
+    """Periodically closes runs whose heartbeat has gone stale.
```
88roy88: remove docstring :)

88roy88: do me a favore and remove from `IdleReaper` as well :)

backend/src/flow44/db/heartbeat.py
```diff
@@ -0,0 +1,141 @@
...
+# ---------------------------------------------------------------------------
+# Heartbeat model
+# ---------------------------------------------------------------------------
+
+
+class AgentRunHeartbeat(SQLModel, table=True):
+    __tablename__ = "agent_run_heartbeat"
+
+    project_id: str = Field(sa_column=Column(String, ForeignKey("projects.id", ondelete="CASCADE"), primary_key=True))
+    beat_at: datetime | None = Field(
```
88roy88: how and why its nullable? (there's a default for now)

backend/src/flow44/db/heartbeat.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def _stale_cutoff() -> datetime:
+    return datetime.now(UTC) - timedelta(seconds=settings.AGENT_RUN_STALE_TIMEOUT)
+
+
+def _stale_or_missing() -> Any:
+    """WHERE clause matching heartbeats that are stale or absent."""
+    beat_at = col(AgentRunHeartbeat.beat_at)
+    return beat_at.is_(None) | (beat_at < _stale_cutoff())
```
88roy88: I dont think here it will be nullable

backend/src/flow44/db/heartbeat.py
```diff
@@ -0,0 +1,141 @@
...
+    await _commit(
+        _insert()(AgentRunHeartbeat)
+        .values(project_id=project_id, beat_at=now)
+        .on_conflict_do_update(index_elements=["project_id"], set_={"beat_at": now})
+    )
+    return now
+
+
+async def try_claim_run(project_id: str) -> datetime | None:
+    """Atomically claim the run-lock.
```
88roy88: remove doc

backend/src/flow44/db/heartbeat.py
```diff
@@ -0,0 +1,141 @@
+import logging
+from datetime import UTC, datetime, timedelta
+from typing import Any
+
+from sqlalchemy import Column, DateTime, ForeignKey, String, delete, func
+from sqlalchemy.dialects.postgresql import insert as pg_insert
+from sqlalchemy.dialects.sqlite import insert as sqlite_insert
```
88roy88, line 7: what do you think, lets open a ticket to remove the sqlite support

88roy88 (changes requested):

backend/src/flow44/db/heartbeat.py
shalom-raichman: I have just encounter a miluimnik that ask if we must use a local postgres, so it has some value for easy onboarding, the trade off is massy code.

backend/src/flow44/services/heartbeat_reaper.py
shalom-raichman: yes sir ;)

backend/src/flow44/db/heartbeat.py
shalom-raichman: it really shouldn't, and corrected

backend/src/flow44/db/heartbeat.py
shalom-raichman: done

backend/src/flow44/db/heartbeat.py
shalom-raichman: done

backend/src/flow44/db/heartbeat.py
shalom-raichman: me personally don't care for the sqlite because I don't use it

88roy88: I think that with our change that s3 for example is a must, now its more reasonalbe to assume that everyone will run the compose and stuff, lets open a ticket about it

shalom-raichman: done

backend/src/flow44/api/chat.py
```diff
@@ -35,20 +37,93 @@ async def _is_new_project(project_id: str) -> bool:
...
+    await emit_event(project_id, {"type": "error", "message": message})
+
+
+async def _run_agent_safe(project_id: str, coro: Any, claimed_at: datetime) -> None:
+    """Run the agent under a heartbeat + timeout, surfacing failures to the client."""
+    run_task = asyncio.create_task(coro)
+    beat = _Beat(at=claimed_at)
     try:
-        await coro
+        await _heartbeat_until_done(project_id, run_task, beat)
```
88roy88, line 96: isnt it opening fir exactly the same issue we had with `await coro`?

88roy88 (changes requested):

backend/src/flow44/api/chat.py
shalom-raichman: claude explanation in short:
_heartbeat_until_done is on _run_agent_safe's stack (kept alive by _RUNNING), and every task it references is held by a live frame. The GC-collection failure mode requires an un-awaited, un-stored create_task — and this line has neither. Nothing to fix

backend/src/flow44/db/project_member.py
```diff
@@ -14,7 +14,9 @@ class ProjectMember(SQLModel, table=True):
     __table_args__ = (UniqueConstraint("project_id", "user_id", name="uq_project_member"),)
 
     id: str = Field(default_factory=lambda: str(uuid.uuid4()), primary_key=True)
-    project_id: str = Field(index=True, foreign_key="projects.id")
+    project_id: str = Field(
```
88roy88, line 17: you added here the cascade correctly but you left the `project_data_source` without it, can you take care of it? (I know its a bit out of scope but since you did it here I want it to be complete)

backend/src/flow44/api/chat.py
```diff
@@ -35,20 +37,93 @@ async def _is_new_project(project_id: str) -> bool:
...
+    """The heartbeat timestamp this run currently owns — its lock token."""
+
+    at: datetime
+
+
+async def _heartbeat_until_done(project_id: str, run_task: asyncio.Task[Any], beat: _Beat) -> None:
+    """Beat while the run executes; raise _RunBudgetExceeded if it overruns its budget."""
+    started = time.monotonic()
+    beat_interval = settings.AGENT_RUN_STALE_TIMEOUT / 4
+    while True:
```
88roy88: I prefer `While started + settings.AGENT_RUN_STALE_TIMEOUT < time.monotonic()` or somethiing similar

backend/src/flow44/api/chat.py
```diff
@@ -35,20 +37,93 @@ async def _is_new_project(project_id: str) -> bool:
...
+        return False
+    task = asyncio.create_task(_run_agent_safe(project_id, coro, claimed_at))
+    _RUNNING.add(task)
+    task.add_done_callback(_RUNNING.discard)
+    return True
 
 
-def _start_agent(project_id: str, coro: Any) -> None:
-    asyncio.create_task(_run_agent_safe(project_id, coro))
+async def _dispatch(websocket: WebSocket, project_id: str, coro: Any) -> None:
```
88roy88: not fan of the _dispatch name here. 
also the logic seems a bit messy, I think that `_start_agent`  should throw if agent is running and not return bool, then you can catch that if you want specific error handling (or even better have a middleware taking care of it)

88roy88 (changes requested):

backend/src/flow44/db/project_member.py
shalom-raichman: done

backend/src/flow44/api/chat.py
shalom-raichman: done

backend/src/flow44/api/chat.py
shalom-raichman: this will be ugly anyway, the function is billion lines and I tried to keep it minimal
the right thing to do is a refactor of the function that do to many thing but it's out of scope
middleware is not possible for mid-run message without closing the connection, which we don't want for a subtle response message.
I agree abut the _dispatch part but I don't really like the other option either

tell what you think, I maybe missing something

backend/src/flow44/db/project_member.py
88roy88: doesnt this require to regenerate the alembic migration file?

88roy88 (approved):

PR #155, opened 2026-06-28
refactor: optimize file tree refresh and remove unused request serial

shalom-raichman:


frontend/src/stores/chatHandlers.ts
```diff
@@ -57,11 +57,17 @@ function generateId(): string {
...
 
+let _treeRefreshTimer: ReturnType<typeof setTimeout> | null = null;
+
+// Agents emit many file/task events in a burst; coalesce them into a single tree refetch.
 function refreshFileTreeAfterAgentWrite() {
   if (_skipMessages) return;
-  const store = useFilesStore.getState();
-  void store.loadFileTree();
   useFilesStore.setState((s) => ({ saveVersion: s.saveVersion + 1 }));
+  if (_treeRefreshTimer) clearTimeout(_treeRefreshTimer);
```
88roy88: can we have 1 debounce hook that will be used in all the different places where we do debounce now (ask claude to find where, I remember more debounce places)

88roy88 (changes requested):

frontend/src/stores/files.ts
```diff
@@ -106,15 +103,14 @@ export const useFilesStore = create<FilesState>((set, get) => ({
       });
     }
 
-    const requestId = ++_fileTreeRequestSerial;
-    const tree = await api.fetchFileTree(projectId);
-    // Drop stale responses (newer loadFileTree started, or project switched).
-    if (requestId !== _fileTreeRequestSerial) return;
-    if (useSessionStore.getState().projectId !== projectId) return;
-    set({ fileTree: tree });
+    const tree = await queryClient.fetchQuery({
```
88roy88: also I want to belive that react query has something for debounce as well

shalom-raichman: React Query (TanStack Query) deliberately doesn't include a built-in debounce function because it focuses strictly on data fetching, caching, and state management. It leaves UI-level concerns—like waiting for a user to stop typing—to you.

frontend/src/stores/chatHandlers.ts
shalom-raichman: I have already wrote a debounce value hook, but it's for the react only so it didn't fit in here and to refactor every debounce here seems excessive, but if we are in the mood of correcting the world, I'm on it

frontend/src/components/editor/useEditorPanelSearch.ts
```diff
@@ -65,32 +65,17 @@ export function useEditorPanelSearch(
...
-      clearTimeout(searchDebounceTimerRef.current);
-    }
+  const debouncedSearch = useDebouncedCallback(() => void performSearch(), 500);
 
-    // Don't auto-search if query is empty
+  useEffect(() => {
     const query = searchQuery.trim();
     if (!query) {
       setSearchResults([]);
       setSearchError(null);
```
88roy88, line 74: [claude] Minor note: clearing the search query no longer cancels the pending debounced search timer. The old code unconditionally cleared `searchDebounceTimerRef` on every effect re-run (including when the query became empty); this version only calls `debouncedSearch()` when the query is non-empty, and takes an early return without `debouncedSearch.cancel()` when it's empty. Right now this is harmless because `performSearch` independently re-checks `searchQuery.trim()` and no-ops on empty, so no incorrect fetch fires — but it's a dropped invariant: if that internal guard is ever removed or reordered, a stale timer could fire an unwanted search after the box was cleared. Consider calling `debouncedSearch.cancel()` in the early-return branch for defense-in-depth.

88roy88 (approved):

PR #158, opened 2026-06-29
Fix unfamiliar field types that cause errors

yoni5353:
Added missing field types "number" and "html", and expanded the literal with `| str`

(the parsing to domain fieldtypes already defaults to "string" when encountering unfamiliar types)


88roy88:
I dont like the `| str` this will catch anything, I prefer to fail and see what we are missing and add them

88roy88 (changes requested):
I dont like the | str this will catch anything, I prefer to fail and see what we are missing and add them.

wdty?

yoni5353:
I did a big ctrl+shift+f and added all types that exist on static fields today

but the problem is that this field can be user-defined in dynamic cubes and flapi doesn't enforce it at all, so there could still be more options i won't be able to find.

so imo we can let this field slide with `| str`, and the `_to_domain_field_type` function already logs a warning if we want to track

88roy88 (approved):

PR #163, opened 2026-07-06
Fix /healthcheck route to actually check backend

yoni5353:


backend/src/flow44/main.py
```diff
@@ -80,6 +78,7 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
 
 
 @app.get("/health")
+@app.get("/api/health")
 async def health_check() -> dict[str, str]:
     """Health check endpoint for Docker/K8s."""
     return {"status": "ok", "version": "0.1.0"}
```
88roy88: A little trick I like for better health endpoint is to return the time as well :)

88roy88 (approved):

backend/src/flow44/main.py
yoni5353: added :-)

PR #164, opened 2026-07-07
Add post-merge alembic validation CI + fix project_data_sources FK cascade

88roy88:
## Summary
- Adds a new GitHub Actions workflow (`alembic-validate.yaml`) that runs on push to `main`/`develop`, spins up a real Postgres service, and verifies: a single alembic migration head (no branch divergence), `alembic upgrade head` applies cleanly, and `alembic check` reports no model/migration drift.
- Fixes `ProjectDataSource.project_id` to declare `ondelete="CASCADE"` on its FK, matching what the migration already set in Postgres — `alembic check` was failing on this drift (verified against a real Postgres container).

Note: `project_members` has a similar cascade gap, already fixed independently in #154.

## Test plan
- [x] Verified locally against `postgres:16-alpine`: single head, `alembic upgrade head` succeeds, `alembic check` passes for `project_data_sources` (remaining `project_members` drift closes once #154 merges)
- [x] `mypy` and `ruff` clean on the changed file
- [x] Full backend test suite passes (350 passed, 66 skipped)


yoni5353:
🙏

yoni5353 (approved):

PR #167, opened 2026-07-16
refactor: remove sqlite support

shalom-raichman:


backend/src/flow44/db/heartbeat.py
```diff
@@ -45,7 +39,7 @@ async def _commit(stmt: Any) -> Any:
 async def touch_heartbeat(project_id: str) -> datetime:
     now = datetime.now(UTC)
     await _commit(
-        _insert()(AgentRunHeartbeat)
+        pg_insert(AgentRunHeartbeat)
```
88roy88: so lets rename to just insert (not import as)

backend/pyproject.toml
```diff
@@ -43,6 +42,8 @@ dev = [
     "pytest-asyncio>=0.23.7,<0.24",
     "pytest-cov>=5.0.0,<6",
     "pytest-xdist>=3.6.1,<4",
+    "pytest-postgresql>=6.0.0",
+    "psycopg[binary]>=3.2.0",
```
88roy88, line 46: is it a dev dependency?

88roy88: doesnt the package name is psycopg2 btw?

88roy88 (approved):

backend/src/flow44/db/heartbeat.py
shalom-raichman: done

backend/pyproject.toml
shalom-raichman: psycopg2 is the legacy version and we use it only for the pytest-postgresql

PR #169, opened 2026-07-19
ADAPI integration for users & groups

orangilboa:


backend/src/alembic/versions/2026_07_09_1000-a7c3e9f10b21_add_project_member_groups_table.py
```diff
@@ -0,0 +1,50 @@
...
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.create_table('project_member_groups',
+    sa.Column('id', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+    sa.Column('project_id', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+    sa.Column('group_id', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+    sa.Column('group_name', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+    sa.Column('role', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+    sa.Column('created_at', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
```
orangilboa, line 34: should be datetime type

backend/src/alembic/versions/2026_07_12_1200-b8d4f2a19c33_add_platform_user_groups_table.py
```diff
@@ -0,0 +1,39 @@
...
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.create_table('platform_user_groups',
+    sa.Column('group_id', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+    sa.Column('group_name', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+    sa.Column('invited_by', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+    sa.Column('created_at', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
```
orangilboa, line 33: DATETIME

backend/src/alembic/versions/2026_07_12_1200-b8d4f2a19c33_add_platform_user_groups_table.py
```diff
@@ -0,0 +1,39 @@
...
+down_revision: Union[str, Sequence[str], None] = 'a7c3e9f10b21'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.create_table('platform_user_groups',
+    sa.Column('group_id', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+    sa.Column('group_name', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
```
orangilboa, line 30: not sure storing group name makes sense, it can change externally (though should be rare)

backend/src/flow44/api/deps.py
```diff
@@ -163,11 +216,35 @@ async def _check(user_permissions: PermissionsDep) -> set[Permission]:
     return Depends(_check)
 
 
+async def _is_platform_group_member(user_id: str) -> bool:
+    """Whether the user belongs to any directory group granted platform access.
+
+    Short-circuits (no ADAPI call) when no group has been granted access, so only
+    deployments that actually use group grants pay the directory round-trip.
+    if this becomes too heavy create an index on group id and use the user's group to search on the db
```
orangilboa: Should this be done right away? Right now the number of groups is 0 and very few users.
On the other hand any given user can be part of dozens or hundreds of groups, DB won't enjoy being bombed with such lists on every login. Update the comment to reflect that thinking

backend/src/flow44/api/deps.py
```diff
@@ -116,13 +124,63 @@ def is_admin(user_id: str) -> bool:
...
     project = await db_get_project(project_id)
     if project is None:
         raise HTTPException(status_code=404, detail="Project not found")
 
     has_access = (
-        project.user_id == user_id or is_admin(user_id) or await get_project_member(project_id, user_id) is not None
+        project.user_id == user_id
+        or is_admin(user_id)
+        or await get_project_member(project_id, user_id) is not None
+        or bool(await resolve_group_permissions(project_id, user_id))
```
orangilboa: seems like a duplicate of of _resolve_project_permissions, validate

backend/src/flow44/api/projects.py
```diff
@@ -60,13 +69,25 @@ def _serialize_project(project: Project, role: str) -> ProjectResponse:
...
 
+# Display precedence when a project reaches a user through more than one source
+# (e.g. a direct share and a group grant). Higher wins; the label is cosmetic —
+# actual authorization unions permissions in deps._resolve_project_permissions.
+_ROLE_RANK: dict[str, int] = {
+    Role.viewer.value: 1,
+    Role.editor.value: 2,
+    Role.publisher.value: 2,
+    Role.maintainer.value: 3,
+    "admin": 4,
```
orangilboa: why are these not under Role.admin & Role.owner as well?

backend/src/flow44/api/projects.py
```diff
@@ -83,15 +104,32 @@ async def list_user_projects(user_id: UserDep) -> list[ProjectResponse]:
...
+
+    def _merge(project: Project, role: str) -> None:
+        existing = by_id.get(project.id)
+        if existing is None or _ROLE_RANK.get(role, 0) > _ROLE_RANK.get(existing[1], 0):
+            by_id[project.id] = (project, role)
 
-    result = []
     for p in owned:
-        result.append(_serialize_project(p, "owner"))
+        _merge(p, "owner")
```
orangilboa: constify

backend/tests/api/test_group_shared_listing.py
```diff
@@ -0,0 +1,84 @@
...
+        assert by_id[project.id]["role"] == Role.editor.value
+
+    async def test_no_groups_omits_group_shared_projects(self, test_db):
+        project = await create_project(name="Team Project", user_id="owner")
+        await add_group(project.id, GROUP_A, "Cloud Leads", Role.editor, "owner")
+
+        with patch(
+            "flow44.api.projects.adapi_client.get_user_group_ids",
+            new=AsyncMock(return_value=set()),
+        ):
```
orangilboa, line 56: I'm not sure this is a meaningful test

backend/tests/api/test_platform_groups.py
```diff
@@ -0,0 +1,88 @@
...
+            assert await has_platform_access("nobody") is False
+        mock_groups.assert_not_awaited()
+
+    async def test_member_of_granted_group_gets_access(self, test_db):
+        await add_platform_group(GUID, "Cloud Leads", "admin-user")
+        with patch(
+            "flow44.api.deps.adapi_client.get_user_group_ids",
+            new=AsyncMock(return_value={GUID}),
+        ):
+            assert await has_platform_access("someone") is True
```
orangilboa: what is the meaning of "someone" here

frontend/src/components/admin/DirectorySearch.tsx
```diff
@@ -0,0 +1,257 @@
...
+    if (next === filter) return;
+    // Invalidate in-flight requests and drop results that fall out of scope.
+    latestSeq.current += 1;
+    setUsers([]);
+    setGroups([]);
+    setFilter(next);
+  };
+
+  const placeholder = useMemo(() => {
+    if (canUsers && canGroups) return t('admin.searchPlaceholder', 'Search users & groups by name or email…');
```
orangilboa, line 132: const unless literals are enforced by type checks

backend/src/flow44/integrations/adapi/client.py
```diff
@@ -0,0 +1,154 @@
+"""Thin client for ADAPI (Active Directory API).
```
88roy88: we dont really like all this AI comments, so in general, remove docstrings and file docs.

backend/src/flow44/integrations/adapi/client.py
```diff
@@ -0,0 +1,154 @@
...
+    """Raised when an ADAPI search cannot be completed (ADAPI down or malformed)."""
+
+
+class _AdRecord(BaseModel):
+    # ADAPI carries more fields than we model; ignore the rest rather than
+    # reject the record. Field names mirror the AD attribute names.
+    model_config = ConfigDict(extra="ignore")
+
+    cn: str
+    displayName: str = ""  # noqa: N815 — mirrors the AD attribute name (wire contract)
```
88roy88: I think we had a base class for those PascalCase or something, 
anyway, a better way to handle is with a `display_name: str = Field(alias="displayName")`

backend/src/flow44/integrations/directory/client.py
```diff
@@ -0,0 +1,154 @@
...
+class AdapiClient:
+    def __init__(
+        self, base_url: str | None = None, *, timeout_s: float | None = None, client_id: str | None = None
+    ) -> None:
+        self.base_url = (base_url or settings.ADAPI_BASE_URL).rstrip("/")
+        self._timeout = timeout_s if timeout_s is not None else settings.ADAPI_TIMEOUT_SECONDS
+        # Sent as the ``ClientId`` header so ADAPI can attribute the call.
+        self._headers = {"ClientId": client_id if client_id is not None else settings.ADAPI_CLIENT_ID}
+
+    async def get_user_group_ids(self, user_id: str) -> set[str]:
```
88roy88, line 50: just for clarity, this user_id is actually a user_email or something, and its not ours internal user_id, maybe rename? wdty?

backend/src/flow44/integrations/adapi/client.py
```diff
@@ -0,0 +1,154 @@
...
+            logger.warning("ADAPI group lookup failed for user %s: %s", user_id, exc)
+            return set()
+
+        match = next((u for u in users if u.mail.casefold() == user_id.casefold()), None)
+        if match is None:
+            logger.warning("ADAPI user not found for group lookup: %s", user_id)
+            return set()
+        return {dn for dn in match.memberOf if dn}
+
+    async def search_users(self, query: str) -> list[AdUser]:
```
88roy88: I think we can call it user_id (or email per the last comment). not just general query, wdyt?

backend/src/flow44/db/project_member_group.py
```diff
@@ -0,0 +1,127 @@
...
+    __table_args__ = (UniqueConstraint("project_id", "group_id", name="uq_project_member_group"),)
+
+    id: str = Field(default_factory=lambda: str(uuid.uuid4()), primary_key=True)
+    project_id: str = Field(
+        sa_column=Column(String, ForeignKey("projects.id", ondelete="CASCADE"), index=True, nullable=False)
+    )
+    # AD distinguishedName (DN) of the group — the identifier ADAPI reports in a
+    # user's ``memberOf``, so access resolution intersects on it directly.
+    group_id: str = Field(index=True)
+    # Human-readable group name, cached for display so the members UI need not
```
88roy88: כמה קריאות לadapi כבדות לנו?
אני מניח שהייתי עושה על זה קאש אבל לא חושב שהייתי שומר דברים כאלה שמשתנים בdb

backend/src/flow44/db/project_member_group.py
```diff
@@ -0,0 +1,127 @@
...
+    group_name: str = Field(default="")
+    # Group email captured at invite time, shown as the row's secondary text under
+    # the group name. Display-only, may go stale — access keys on group_id/DN.
+    email: str = Field(default="")
+    role: str = Field(default=Role.viewer.value)
+    created_at: datetime | None = Field(
+        default=None,
+        sa_column=Column(DateTime(timezone=True), server_default=func.now(), nullable=False),
+    )
+    invited_by: str = Field(default="")
```
88roy88, line 30: אנחנו לא רוצים foreign key על יוזר?

backend/src/flow44/db/platform_user.py
```diff
@@ -9,12 +10,21 @@ class PlatformUser(SQLModel, table=True):
     __tablename__ = "platform_users"
 
     user_id: str = Field(primary_key=True)
+    # ADAPI displayName captured at invite time, shown as the row's primary text
+    # (the email in user_id is the secondary text). Display-only, may go stale.
+    display_name: str = Field(default="")
```
88roy88, line 14: I guess this will get stale much less often

backend/src/alembic/versions/2026_07_19_1400-d1a2b3c4e5f6_add_display_name_and_group_email.py
```diff
@@ -0,0 +1,61 @@
+"""add display_name to users and email to group grants
```
88roy88: we have the benefit now as we are in the future :) to edit the 2 originals and add those fields so it wont be an additional separate migration

88roy88 (changes requested):

backend/src/flow44/integrations/adapi/client.py
orangilboa: It's a very weird convention imo as these comments mostly help ai reduce hallucinations and read less code to understand the purpose of files/classes/functions

backend/src/flow44/integrations/directory/client.py
orangilboa: Yep. I wouldn't rename because it does serve as user id, I'll make sure it's documented. For most devs and future features it's not interesting anyway

backend/src/flow44/integrations/adapi/client.py
orangilboa: It can also be the user's (or group's) display name, and maybe other fields in the future

backend/src/flow44/db/project_member_group.py
orangilboa: Wasn't sure myself about it, but my reasoning is that display names change seldomly, and it doesn't justify caching on its own. So another column felt like a reasonable compromise.
Adapi is quite slow (1-5s)

backend/src/flow44/db/project_member_group.py
orangilboa: Not as long as it's audit only. If I remove a user I don't wanna remove all the users they invited

backend/src/flow44/integrations/adapi/client.py
orangilboa: I'll look for it

backend/src/alembic/versions/2026_07_19_1400-d1a2b3c4e5f6_add_display_name_and_group_email.py
orangilboa: wouldn't create a mess for devs already working with a local db?
This is kinda the core issue with db client like alembic that have no support for auto-migrations

backend/src/flow44/integrations/adapi/client.py
orangilboa: removed almost all comments

backend/src/flow44/integrations/adapi/client.py
orangilboa: modified to use the Field class

backend/src/alembic/versions/2026_07_19_1400-d1a2b3c4e5f6_add_display_name_and_group_email.py
orangilboa: changed

backend/src/flow44/db/project_member_group.py
88roy88: no one said to add a cascade :) and I dont think we have a user delete at all.
fair enough

backend/src/alembic/versions/2026_07_19_1400-d1a2b3c4e5f6_add_display_name_and_group_email.py
88roy88: developers in this case == you :)

backend/src/flow44/integrations/adapi/client.py
88roy88: in general im not a fan of the whole implicit query stuff (the whole _search_params that is taking everything and everything). its like we have a huge `search_by(*, text)`.

backend/src/alembic/versions/2026_07_19_1400-d1a2b3c4e5f6_add_display_name_and_group_email.py
88roy88: oh we misunderstood one another, I ment that since
```
   op.add_column(
        'project_member_groups',
        sa.Column('email', sqlmodel.sql.sqltypes.AutoString(), nullable=False, server_default=''),
    )
    op.add_column(
        'platform_user_groups',
        sa.Column('email', sqlmodel.sql.sqltypes.AutoString(), nullable=False, server_default=''),
    )
 ```
are new you can just add this in the migration that create those tables not after the fact.

you are 100% right, we dont touch old, already merged, migration files

File: backend/src/alembic/versions/2026_06_08_1026-bf979c9b9417_add_platform_users_and_project_members_.py [hunk 19]
```diff
@@ -24,6 +24,7 @@ def upgrade() -> None:
     # ### commands auto generated by Alembic - please adjust! ###
     op.create_table('platform_users',
     sa.Column('user_id', sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+    sa.Column('display_name', sqlmodel.sql.sqltypes.AutoString(), nullable=False, server_default=''),
```
88roy88: yeah, no. sorry for the confusion

backend/src/flow44/api/deps.py
```diff
@@ -163,11 +205,26 @@ async def _check(user_permissions: PermissionsDep) -> set[Permission]:
     return Depends(_check)
 
 
+async def _is_platform_group_member(user_id: str) -> bool:
+    # Fetch the (small) set of granted group DNs and intersect in memory against
```
88roy88: not sure I see the points in this kind of ai comments

backend/src/flow44/api/deps.py
```diff
@@ -116,15 +123,60 @@ def is_admin(user_id: str) -> bool:
...
+
+    permissions |= await resolve_group_permissions(project.id, user_id)
+    return permissions
+
+
+# Per-request cache: get_project and the permission dependencies both resolve the
+# same (project, user) set, so resolving once avoids a second member lookup and
+# ADAPI round-trip. ContextVars are per-task, so this never leaks across requests;
+# the key guards against reuse for a different project/user in the same task.
+_perm_cache: ContextVar[tuple[tuple[str, str], set[Permission]] | None] = ContextVar("_perm_cache", default=None)
```
88roy88: cant we just use some lrucache decorator? the different tasks inputs will solve it for us

backend/src/flow44/api/projects.py
```diff
@@ -79,19 +103,33 @@ async def list_user_projects(user_id: UserDep) -> list[ProjectResponse]:
         all_projects = await list_all_projects()
         result: list[ProjectResponse] = []
         for p in all_projects:
-            role = "owner" if p.user_id == user_id else "admin"
```
88roy88, line 82: lets use the enum value directly?

backend/src/flow44/api/projects.py
```diff
@@ -79,19 +103,33 @@ async def list_user_projects(user_id: UserDep) -> list[ProjectResponse]:
...
 
-    owned, shared = await asyncio.gather(db_list_user_projects(user_id), list_shared_projects(user_id))
+    owned, shared, user_group_ids = await asyncio.gather(
+        db_list_user_projects(user_id),
+        list_shared_projects(user_id),
+        adapi_client.get_user_group_ids(user_id),
+    )
+    group_shared = await list_group_shared_projects(user_group_ids)
+
+    # Collapse to one entry per project, keeping the highest-ranked role for display.
```
88roy88, line 101: not sure ranking is the right thing here, I think publisher and writer gives you different operations you can do.

88roy88 (changes requested):

backend/src/flow44/api/projects.py
orangilboa: wdym? using literals was wrong so it was changed to consts

backend/src/flow44/api/deps.py
orangilboa: I think it explains the reasoning for a logic that would otherwise raise questions. removed

backend/src/flow44/api/projects.py
orangilboa: you're right but this ranking is cosmetic in this case, I think associating multiple roles per project per user is out of scope for this task

backend/src/flow44/api/deps.py
orangilboa: no, I don't want a process-level memoization, I want it to live per request and removed when the request is complete

backend/src/flow44/integrations/adapi/client.py
orangilboa: I disagree with that logic, a user submits a query that is bound to certain fields today. It may be different fields tomorrow. Changed the param name to specify which fields are being queried.

backend/src/flow44/integrations/adapi/client.py
```diff
@@ -60,8 +60,8 @@ async def get_user_group_ids(self, user_id: str) -> set[str]:
             return set()
         return {dn for dn in match.member_of if dn}
 
-    async def search_users(self, query: str) -> list[AdUser]:
-        return await self._search_users(self._fuzzy_params(query))
+    async def search_users(self, accountOrEmailOrName: str) -> list[AdUser]:
```
88roy88: `accountOrEmailOrName` -> `account_or_email_or_name`

88roy88 (changes requested):
linter is not passing

backend/src/flow44/integrations/adapi/client.py
88roy88: this is the client, the client should be very explicit, later, if the endpoint level would want to allow a search that is fuzzy and can find by email and name and whatever, we can support it. but here, I think we should be more explicit

backend/src/flow44/api/projects.py
88roy88: we have `class Role(StrEnum):`
maybe owner and admin should be there as well and then we could use
Role.Owner here instead of a const that is sorta duplication? 
(or at least have those const in the same file as the `class Role(StrEnum):` stuff).

also, and i know its not you but existing code (probably even by me :) )
if we do end up using Role class here, can we change the function sig `ProjectResponse` to use the class instead of str?

backend/src/flow44/api/projects.py
88roy88: sababa

backend/src/alembic/versions/2026_06_08_1026-bf979c9b9417_add_platform_users_and_project_members_.py
88roy88: what about this and [this](https://github.com/88roy88/FlowBolt/pull/169/files/42b256652e8027b74e0c82d0459a096eb002ac0a#diff-576e6bb9ec2ffcdcdab62b6f78ca9ea063ce7779f63d18abb71988db2ed1d327)?

(please dont resolve my comments, this is my way of keeping track of stuff)

backend/src/flow44/api/deps.py
88roy88: ok, I hear ya, what about 
`request.state.perm_cache` to use a more native fastapi way for this then? 
and be explicit about request level as well :)

backend/src/flow44/integrations/adapi/client.py
orangilboa: changed according to your request, please resolve

backend/src/alembic/versions/2026_07_19_1400-d1a2b3c4e5f6_add_display_name_and_group_email.py
orangilboa: these were moved to the migrations that create the tables

backend/src/flow44/api/deps.py
orangilboa: I think it adds some complexity for small benefit, and is also slightly out of scope. Roles & permissions should be it's own separate task & PR, not some "by the way" changes to an ADAPI & groups integration task. 
I prefer the MVP attitude for roles & permissions in this PR.

It would also mean we need to pass and handle Request | Websocket in all calls for the same behavior.

However it's not acute imo, I can make the change if you still think I should

orangilboa: I'll change the name to be more explicit

backend/src/flow44/api/projects.py
orangilboa: imo this as well should be done as part of a dedicated roles & permissions task.
Moving the consts to the file

backend/src/alembic/versions/2026_06_08_1026-bf979c9b9417_add_platform_users_and_project_members_.py
orangilboa: addressed

88roy88 (approved):
A few things to address in follow-up or before deploy:

---

**1. `client.py` — search_users should accept explicit params**

The client uses a single `account_or_email_or_name: str` that gets fuzzy-matched against multiple fields. The client should be explicit about what it's searching — separate `account`, `email`, `name` params that build the query precisely. The endpoint layer can then decide to expose a combined fuzzy search if needed, but the client shouldn't hide which fields are being hit.

---

**2. LDAP filter injection in `_fuzzy_params` and `get_user_group_ids`** (`client.py`)

User input from the `/api/adapi/users?q=...` endpoint flows directly into `_fuzzy_params` which builds LDAP filters like `(displayName=*{query}*)` without escaping. LDAP metacharacters (`*`, `(`, `)`, `\`, NUL) are not sanitized.

A user can send `q=*)(objectClass=*` and the filter becomes `(displayName=**)(objectClass=**)` — matching everything in the directory.

The `get_user_group_ids` path has the same pattern with `(mail={user_id})` — lower risk since user_id comes from JWT, but parentheses are valid in email local-parts per RFC 5321.

Fix: add an `_escape_ldap(value: str)` helper per RFC 4515 and apply it before interpolation.

---

**3. Empty query dumps the entire directory** (`adapi.py`)

`q` defaults to `""`. `_fuzzy_params("")` produces `**` which is a wildcard matching all records. Any authenticated user can call `/api/adapi/users` with no params and get back the full corporate directory.

Fix: add `min_length=1` to the Query validator, or return `[]` when `q` is empty.

---

**4. ADAPI failure silently removes group permissions** (`deps.py`)

If ADAPI is unreachable, `get_user_group_ids` returns `set()` — users silently lose all group-based access with no indication. Should either return 503 or log at WARNING level so it's visible in monitoring.

---

**5. `group_id` path param breaks on DNs with `/`** (`members.py`, `admin.py`)

Group IDs are Distinguished Names which can legally contain `/`. FastAPI's default `{group_id}` path type won't match these. Need `{group_id:path}` or switch to a query param.

---

**6. New httpx client created per request** (`client.py`)

Each `_search` call creates and destroys an `httpx.AsyncClient`, paying TCP+TLS setup every time. ADAPI is already slow (1-5s) — connection reuse would help. Create a persistent client at module level.

---

**7. Unbounded ADAPI call on every project listing** (`projects.py`)

Every non-admin project list hits ADAPI with no caching. Combined with the per-request client creation (#6), this adds 1-5s latency to a frequently called endpoint. Consider a short TTL cache on `get_user_group_ids` per user.

---

**8. created_at as VARCHAR then cast to timestamptz** (migrations)

The initial migration creates `created_at` as `AutoString`. The fix migration casts via `::timestamptz`. Any rows inserted between the two migrations running with non-ISO values will cause the cast to fail. Since both migrations ship together in this PR it's likely fine, but worth noting.

backend/pnpm-project-template/pnpm-lock.yaml
yoni5353: why was this deleted..

backend/src/flow44/api/admin.py
```diff
@@ -91,6 +120,48 @@ async def revoke_user(user_id: AdminDep, target_user_id: str) -> None:
         raise HTTPException(status_code=404, detail="User not found")
 
 
+# --- Platform group grants (grant platform access to a whole directory group) ---
```
yoni5353: the comment is too specific, it feels like. these are general for groups

.gitignore
```diff
@@ -67,3 +67,5 @@ AI_INSTRUCTIONS.md
 .agent
 .antigravity
 .antigravityignore
+
+*pnpm-lock.yaml
```
yoni5353: why??

backend/src/flow44/api/deps.py
```diff
@@ -100,13 +106,14 @@ def validate_token(token: TokenDep) -> TokenPayload:
 
 
 def get_user_id(token: TokenDep) -> str:
-    """Resolve a token to a user_id; a valid signed JWT with a ``/UniqueID`` claim is required."""
+    """Resolve a token to a user_id; a valid signed JWT with an ``/emailaddress`` claim is required."""
```
yoni5353: we talked about reverting this. no reason the identifier should be the mail and the not user's id. 

also the relevant function names and parameters were not changed completely anyway (this function is still named get_user_id, for instance...)

backend/src/flow44/config.py
```diff
@@ -98,6 +98,15 @@ class FlapiSettings(BuildAppSettings):
     FLAPI_VERIFY_SSL: bool = True
 
 
+class DirectorySettings(BuildAppSettings):
+    # Defaults to the local directory-mock in dev.
+    DIRECTORY_BASE_URL: str = "http://localhost:6666"
+    DIRECTORY_VERIFY_SSL: bool = True
+    DIRECTORY_TIMEOUT_SECONDS: float = 5.0
```
yoni5353: unfortunately, this feels to me like a too strict timeout, in my experience searches can take a bit more time
```suggestion
    DIRECTORY_TIMEOUT_SECONDS: float = 10.0
```

backend/src/flow44/integrations/directory/client.py
```diff
@@ -0,0 +1,110 @@
...
+        timeout_s: float | None = None,
+        client_id: str | None = None,
+    ) -> None:
+        self.base_url = (base_url or settings.DIRECTORY_BASE_URL).rstrip("/")
+        self._timeout = timeout_s if timeout_s is not None else settings.DIRECTORY_TIMEOUT_SECONDS
+        self._headers = {"ClientId": client_id if client_id is not None else settings.DIRECTORY_CLIENT_ID}
+
+    async def get_user_group_ids(self, user_id: str) -> set[str]:
+        try:
+            users = await self._search_users({"customFilter": f"(mail={user_id})"})
```
yoni5353: change this to the `{aNR: }` query that should be a lot quicker

backend/src/flow44/integrations/directory/client.py
```diff
@@ -0,0 +1,110 @@
...
+        return {dn for dn in match.member_of if dn}
+
+    async def search_users(self, account_or_email_or_name: str) -> list[AdUser]:
+        return await self._search_users(self._fuzzy_params(account_or_email_or_name))
+
+    async def _search_users(self, params: dict[str, str]) -> list[AdUser]:
+        return [AdUser.model_validate(r) for r in await self._search("/users", params)]
+
+    async def search_groups(self, query: str) -> list[AdGroup]:
+        return [AdGroup.model_validate(r) for r in await self._search("/groups", self._fuzzy_params(query))]
```
yoni5353: if in not mistaken the query for groups should only find security groups, so the filter should be added to this requet

PR #128, opened 2026-06-11
Some UI adjustments

yoni5353:
Tone down the blue colors, fix hebrew text, add optimistic update to project deletion

before:
image

after: (notice project corners not cut off; input background color; backgrounds are lighter (admittedly it's still a bit too teal but white looked weird too idk)
image

PR #170, opened 2026-07-20
Migrate observability from Langfuse to Opik

88roy88:
## Summary
- Replace Langfuse with Opik across all agents (plan, execute, follow-up, fix-error): traces tag/thread by project id, manual execute-agent spans nest correctly under their parent step instead of flattening to the trace root, and LLM spans carry `available_tools` (name + description) and `previous_tool_calls` (tool/args/result of the prior iteration) for easier debugging in the Opik UI.
- Bump litellm past its old `<1.82.3` pin now that Opik's own version constraints (`!=1.81.*,!=1.82.*,...`) allow a newer resolution (1.80.17 → 1.93.0).
- Revert an unrelated PR (#128) that had swapped user/assistant chat bubble sides — restores user-right/assistant-left.

## Test plan
- [x] `mypy` / `ruff check` / `ruff format --check` all clean
- [x] Full backend test suite passes (496 tests)
- [x] Verified live against Opik Cloud: ran build → follow-up → fix-error flows end-to-end through the actual app UI, confirmed traces/spans land correctly with proper nesting, project tags, and tool-visibility metadata via the Opik API

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,12 +205,10 @@ Edit `AIB_AI_MODEL` and `AIB_AI_BASE_URL` in `backend/.env` to switch providers:
 
 ## Observability (optional)
 
-Set the following in `backend/.env` to enable Langfuse tracing:
+Set the following in `backend/.env` to enable Opik tracing:
 
 ```env
-AIB_LANGFUSE_PUBLIC_KEY=pk-lf-...
-AIB_LANGFUSE_SECRET_KEY=sk-lf-...
-AIB_LANGFUSE_HOST=https://cloud.langfuse.com
+AIB_OPIK_API_KEY=...
+AIB_OPIK_WORKSPACE=your-workspace
+AIB_OPIK_PROJECT_NAME=flow44
 ```
-
-A self-hosted Langfuse stack is available under `mocks/langfuse/docker-compose.yaml`.
--- a/backend/src/flow44/ai/agents/_base.py
+++ b/backend/src/flow44/ai/agents/_base.py
@@ -1,6 +1,8 @@
+import traceback as traceback_module
 from typing import Any
 
-from langfuse.decorators import langfuse_context
+from opik import opik_context
+from opik.types import ErrorInfoDict
 
 from flow44.db.events import emit_event
 from flow44.sandbox.main import PnpmSandbox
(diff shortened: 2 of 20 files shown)


backend/src/flow44/ai/core/react_flow.py
```diff
@@ -67,11 +67,28 @@ async def react_loop(
...
 
         for iteration in range(self.max_iterations):
             # Call LLM with tools
-            metadata = metadata_fn(f"react-{iteration}") if metadata_fn else None
+            metadata = (
+                metadata_fn(
+                    f"react-{iteration}",
+                    extra_metadata={
+                        "available_tools": available_tools,
+                        "previous_tool_calls": list(last_tool_calls.values()) or None,
```
88roy88: I dont think this is needed anymore

backend/src/flow44/ai/agents/_base.py
```diff
@@ -25,22 +40,76 @@ def __init__(
...
         await emit_event(self.project_id, event)
 
-    def _llm_metadata(self, generation_name: str) -> dict[str, Any]:
-        trace_id = self._trace_id or langfuse_context.get_current_trace_id()
-        observation_id = langfuse_context.get_current_observation_id()
-        return {
-            "existing_trace_id": trace_id,
-            "parent_observation_id": observation_id,
+    def _set_trace_output(self, output: dict[str, Any]) -> None:
+        """Record a summary on the top-level trace so its outcome is visible without opening spans."""
```
88roy88: remove all docstrings

backend/src/flow44/ai/agents/_base.py
```diff
@@ -25,22 +40,76 @@ def __init__(
...
+        if opik_context.get_current_trace_data() is not None:
+            opik_context.update_current_trace(output=output)
+
+    @staticmethod
+    def _safe_span(opik_client: Any, **kwargs: Any) -> Any:
```
88roy88: not a fan, of this logic in general and of returning any here

88roy88: maybe lets move all those opik help functions to the opik. failour logger or something? I dont think they belongs here

backend/src/flow44/ai/core/opik_failure_logger.py
```diff
@@ -0,0 +1,67 @@
...
+        self,
+        kwargs: dict[str, Any],
+        response_obj: Any,
+        start_time: datetime,
+        end_time: datetime,
+    ) -> None:
+        self._log_failure(kwargs, start_time, end_time)
+
+    def _log_failure(self, kwargs: dict[str, Any], start_time: datetime, end_time: datetime) -> None:
+        if self._opik_client is None:
```
88roy88: not sure I see how it can be none, we also have this init function that we later discuessed about moving, maybe we need to override it here, so its always non null and always initilaized (here in this class?)

backend/src/flow44/ai/core/react_flow.py
88roy88: Fixed in 9538840 — removed `previous_tool_calls` / `last_tool_calls` entirely. The full LLM input (which Opik already captures) contains the same info. Only `available_tools` stays as span metadata.

backend/src/flow44/ai/agents/_base.py
88roy88: Done in 9538840 — all docstrings removed from `_base.py`.

backend/src/flow44/ai/agents/_base.py
88roy88: Fixed in 9538840 — dropped the null-object fallback entirely. `create_span()` now calls `get_global_client().span()` directly, returns the real `opik.Span` type (not `Any`), and lets exceptions propagate if Opik is unreachable (fails loudly instead of silently producing untraced runs). `_NullSpan` class is gone.

88roy88: Done in 9538840 — moved all Opik helpers (`setup_trace`, `set_trace_output`, `create_span`, `record_span_error`, `error_info`, `llm_metadata`) into `opik_failure_logger.py` as free functions. `BaseAgent` is now a thin delegator with no Opik-specific logic.

backend/src/flow44/ai/core/opik_failure_logger.py
88roy88: Fixed in 9538840 — `FailureAwareOpikLogger.__init__` now overrides to call `self._opik_client = get_global_client()` directly. Since `opik` is a hard dependency (not optional), `get_global_client()` always returns a real `Opik` instance — the `None` check and its unreachable-code guard are gone.

README.md
```diff
@@ -205,12 +205,10 @@ Edit `AIB_AI_MODEL` and `AIB_AI_BASE_URL` in `backend/.env` to switch providers:
...
 ```env
-AIB_LANGFUSE_PUBLIC_KEY=pk-lf-...
-AIB_LANGFUSE_SECRET_KEY=sk-lf-...
-AIB_LANGFUSE_HOST=https://cloud.langfuse.com
+AIB_OPIK_API_KEY=...
+AIB_OPIK_WORKSPACE=your-workspace
+AIB_OPIK_PROJECT_NAME=flow44
 ```
-
-A self-hosted Langfuse stack is available under `mocks/langfuse/docker-compose.yaml`.
```
yoni5353, line 216: so let's delete the langfuse mock from `/mocks`

backend/src/flow44/ai/agents/plan/agent.py
```diff
@@ -205,9 +208,11 @@ async def rebuild_with_feedback(self, state: BuildState, feedback: str) -> None:
         await self.emit({"type": "phase", "phase": "awaiting_approval"})
         await self.emit({"type": "plan_overview", "overview": self._state.user_overview.model_dump()})
 
+        self._set_trace_output({"user_overview": self._state.user_overview.model_dump()})
```
yoni5353: just my opinion but `user-overview` is a weird shorthand for `user-plan-overview`, it think it would have been much clearer as `plan-overview` or `user-plan-overview`. At first glance i thought it was info about the user or something

same for the `@track(name="build-user-overview")` step in line 243

yoni5353 (approved):
looks good, left minor comments for you to consider

README.md
yoni5353: nvm i saw you removed it in the next pr

backend/src/flow44/ai/agents/plan/agent.py
88roy88: cool, agree, 
heard you tested it and it worked with the infra.
please apply the above ^ and merge it.
tnx <3

PR #171, opened 2026-07-20
Replay real tool-call history on follow-up requests

88roy88:
## Summary
Stacked on #170 (Opik migration) — targets that branch, not `develop`, until it merges.

- `FollowUpAgent` previously dropped prior `tool_call`/`tool_result`/`reasoning` rows when rebuilding conversation history for the LLM on a new follow-up, keeping only user messages and final assistant answers — so the model (and Opik traces) never saw what tools ran on earlier turns.
- Adds a `raw_message` JSON column to `chat_messages` so `ReActFlow`'s real assistant-with-`tool_calls` and tool-role messages get persisted and replayed verbatim on the next follow-up. The model sees what it actually did, not a paraphrased summary.
- Tool-call arguments and results are truncated (500/200 chars) to keep prompt growth bounded over a long conversation — full fidelity everywhere else.
- Legacy rows saved before this change (no `raw_message`) fall back to the old plain-text folding, so existing conversations don't break.

## Test plan
- [x] `mypy` / `ruff check` / `ruff format --check` all clean
- [x] Full backend test suite passes (496 tests)
- [x] `alembic check` — no drift, migration applies cleanly
- [x] Verified reconstructed history round-trips through `litellm.acompletion` without error (valid message shape)
- [x] Verified against a real project's chat history: reconstructed messages alternate real `assistant` (with `tool_calls`) / `tool` (matched by `tool_call_id`) pairs exactly matching a live ReAct loop's shape


backend/src/flow44/main.py
```diff
@@ -1,15 +1,12 @@
...
 from contextlib import asynccontextmanager
 from datetime import UTC, datetime
 
-import litellm
-import opik
 from fastapi import APIRouter, Depends, FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 
-from flow44.ai.core.opik_failure_logger import FailureAwareOpikLogger
+from flow44.ai.core.provider import flush_opik_traces, setup_opik_tracing
```
88roy88: move to opik_failure_logger and rename it maybe to opik_utils or something

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -173,11 +173,7 @@ async def run(self, content: str, data_source_ids: list[str] | None = None) -> N
             self._build_context(),
             get_messages(self.project_id),
         )
-        messages = [
-            Message(role=m.role, content=m.content)
-            for m in history
-            if m.role == "user" or (m.role == "assistant" and m.content.strip())
-        ]
+        messages = self._history_to_messages(history)
```
88roy88, line 176: and where is the last message? the one that started this follow up agent?

backend/src/flow44/ai/core/react_flow.py
```diff
@@ -17,6 +17,20 @@
 
 StateT = TypeVar("StateT", bound=BaseModel)
 
+MAX_ARG_LENGTH = 500
+
+
+def _truncate_args(args: dict[str, Any]) -> dict[str, Any]:
```
88roy88: does this should be here or in the Tool class file?

backend/src/flow44/ai/core/react_flow.py
```diff
@@ -44,7 +58,7 @@ def __init__(
 
     async def react_loop(
         self,
-        messages: list[Message],
+        messages: list[dict[str, Any] | Message],
```
88roy88, line 30: why the history messages are not being converted to list[Message] as well?

backend/src/flow44/ai/core/react_flow.py
```diff
@@ -104,24 +120,38 @@ async def react_loop(
...
-            if emit_fn and interim_reasoning:
+            # Add assistant message with tool calls
+            assistant_message = message.model_dump()
+            working_messages.append(assistant_message)
+
+            # The real assistant turn (content + reasoning_content + tool_calls), with any large
+            # tool-call argument values truncated — this is what gets replayed verbatim as history
+            # on future follow-ups, so it must not carry unbounded file content forever.
+            raw_assistant_message = dict(assistant_message)
+            raw_assistant_message["tool_calls"] = [
```
88roy88, line 83: same, lets have something in Tool for it?

backend/src/flow44/ai/core/react_flow.py
```diff
@@ -104,24 +120,38 @@ async def react_loop(
             choice = response.choices[0]
             message = choice.message
             last_content = message.content or ""
-            reasoning_content = getattr(message, "reasoning_content", None) or ""
```
88roy88, line 107: not sure I get this reasoning_content / working_messages at al tbh

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -259,13 +285,15 @@ async def _build_context(self) -> dict[str, Any]:
 
     async def _emit_react_step(self, event: dict[str, Any]) -> None:
         """Emit ReAct step events and track state for followup agent."""
-        if event["type"] == "react_reasoning":
+        if event["type"] == "react_assistant_turn":
```
88roy88, line 285: what is this event type? when do we have it? why do we have it? is it used?

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -259,13 +285,15 @@ async def _build_context(self) -> dict[str, Any]:
...
+            # The real assistant message (content + reasoning_content + tool_calls) for this
```
88roy88: I said no docstring, this means also no AI comments like this.

backend/src/flow44/ai/agents/followup/agent.py
```diff
@@ -227,6 +223,36 @@ async def _write_data_source_module(self, ctx: DataSourceContext) -> None:
...
+
+        for m in history:
+            if m.role == ChatRole.user:
+                messages.append(Message.user(m.content))
+            elif m.raw_message is not None:
+                messages.append(m.raw_message)
+            elif m.role == ChatRole.assistant and m.content.strip():
+                messages.append(Message.assistant(m.content))
+            else:
+                has_legacy = True
```
88roy88: I think it shouldnt be binary, we should have all the messages from the user until last legacy to be truncut and then start accumelate the good ones. 
also, do you think we are missing a lot by not using the other user messages / assistant messages we do have?

backend/src/flow44/main.py
88roy88: Done — renamed to `opik_utils.py`, moved `setup_opik_tracing`/`flush_opik_traces` there from `provider.py`. `main.py` imports from `opik_utils` now.

backend/src/flow44/ai/agents/followup/agent.py
88roy88: It's the last `ChatRole.user` row in the DB — `chat.py`'s WebSocket handler saves the user message (`save_message(project.id, ChatRole.user, user_content)`) *before* calling the followup agent. So `get_messages()` already includes it as the final row by the time `_history_to_messages` runs.

backend/src/flow44/ai/core/react_flow.py
88roy88: Moved to `tools.py` as `truncate_tool_args()`.

backend/src/flow44/ai/core/react_flow.py
88roy88: History messages come from `raw_message` which are plain dicts (the literal LLM message structure with `role`/`content`/`tool_calls`/`tool_call_id`). They can't be `Message` instances because `Message` doesn't model tool-role messages or the `tool_calls` array on assistant messages — it's a simpler wrapper for user/assistant/system text. The `list[dict | Message]` union is intentional: new user messages are `Message`, replayed history is raw dicts passed through as-is.

backend/src/flow44/ai/core/react_flow.py
88roy88: Moved to `tools.py` alongside `truncate_tool_args`. The truncation builds a shortened copy of the assistant message for DB storage — it's tool-argument-level logic.

backend/src/flow44/ai/core/react_flow.py
88roy88: Removed the `reasoning_content` extraction block entirely (it was a leftover from the old `react_reasoning` event — that event type was removed earlier). `working_messages` is just the running LLM conversation array that gets passed to each iteration's `complete_chat_with_tools` call.

backend/src/flow44/ai/agents/followup/agent.py
88roy88: `react_assistant_turn` is emitted by `react_flow.py` once per LLM iteration — it carries the full assistant message (with tool_calls, truncated args) so the followup agent can save it to the DB as a replayable `raw_message`. Without it, we'd only have the per-tool `react_step` events which don't carry the assistant-turn-level structure needed for history replay.

backend/src/flow44/ai/agents/followup/agent.py
88roy88: Removed — all docstrings and comments stripped from `_emit_react_step` and `react_flow.py`.

backend/src/flow44/ai/agents/followup/agent.py
88roy88: Fixed — no longer binary. Now walks through all rows in order: user/assistant/raw_message rows are kept as-is, and only legacy tool_call/tool_result rows (that have no `raw_message`) get compacted into a single synthetic assistant message *between* the user messages that bookend them. So all real user turns and assistant answers are preserved in the prompt.

backend/src/flow44/ai/core/opik_utils.py
```diff
@@ -121,3 +126,18 @@ def llm_metadata(
...
+        logger.info("Opik tracing not enabled (missing API key)")
+        return
+    litellm.callbacks = [FailureAwareOpikLogger()]
+    logger.info("Opik tracing enabled")
+
+
+async def flush_opik_traces() -> None:
+    if not settings.OPIK_API_KEY:
+        return
+    await asyncio.to_thread(opik.flush_tracker)
```
88roy88: whats the point of await to asyncio.to_thread?

backend/src/flow44/ai/core/opik_utils.py
```diff
@@ -121,3 +126,18 @@ def llm_metadata(
...
+
+def setup_opik_tracing() -> None:
+    if not settings.OPIK_API_KEY:
+        logger.info("Opik tracing not enabled (missing API key)")
+        return
+    litellm.callbacks = [FailureAwareOpikLogger()]
+    logger.info("Opik tracing enabled")
+
+
+async def flush_opik_traces() -> None:
```
88roy88: why this async

88roy88: Dropped — made it a plain sync function. It runs during shutdown after all requests are done, blocking briefly is fine.

backend/src/flow44/ai/core/opik_utils.py
88roy88: Removed — `opik.flush_tracker()` is blocking (polls with `time.sleep`) but since we only call it during lifespan shutdown (no more requests to serve), there's no reason to push it to a thread. Simplified to a direct sync call.

yoni5353 (approved):

PR #172, opened 2026-07-20
Opik cleanup: provider.py refactor, drop mocks/langfuse, legacy history migration

88roy88:
## Summary
Stacked on #171 (roym/followup-tool-call-history). Cleanup follow-through from the Opik migration:

- Move litellm/Opik callback setup + shutdown flush from `main.py` into `provider.py`, alongside the rest of the LLM provider code.
- Delete the dead `mocks/langfuse/` docker-compose + Helm chart — nothing in the repo references it anymore.
- Backfill `raw_message` on legacy `chat_messages` rows (best-effort reconstruction from the old summary-string format) and drop the now-dead `reasoning` role. `FixErrorAgent`'s synthetic tool-call step now synthesizes `raw_message` going forward too, so the legacy text-folding fallback in `_history_to_messages` could be deleted outright instead of kept around indefinitely.
- Stop threading the Opik client through `ExecutionState` — `_safe_span` now calls the cheap, cached `get_global_client()` directly instead of every caller fetching and passing it through Flow state.
- Also includes a small unrelated fix: replace deprecated Pydantic class-based `Config` with `ConfigDict` in `execution_state.py`/`fix_error_state.py`/`plan_state.py`.

## Test plan
- [x] `uv run ruff check` / `uv run ruff format --check` — all pass
- [x] `uv run mypy --config-file mypy.toml src/` — no issues
- [x] `uv run pytest -q --no-cov` — 496 passed
- [x] Manually ran the alembic migration's `upgrade()`/`downgrade()` against a synthetic sqlite copy of the dev DB to verify reasoning-folding, tool_call_id pairing, and raw_message nulling on downgrade


88roy88:
Merged into #171 — all commits cherry-picked onto roym/followup-tool-call-history.

PR #175, opened 2026-07-26
feat: serve published apps from real S3 dist tree via /shared proxy

shalom-raichman:
- deploy the full Vite dist/ to S3 and proxy every asset (ETag/304), replacing the single inlined HTML — fixes 404s on images, JS/CSS, fonts
- extract publish/shared logic into services/; keep routers thin
- fix export HTML/ZIP 500 on non-ASCII (Hebrew) project names (RFC 5987)


backend/src/flow44/api/export.py
```diff
@@ -21,6 +22,13 @@
 EXCLUDED_DIRS = {"node_modules", ".git", "dist", ".cache"}
 
 
+def _download_as(filename: str) -> str:
```
88roy88: this isnt really download as, unless you return the whole response here,
make a or `def prepare_download(content, name) -> Response` even `class DownloadContent` or

backend/src/flow44/sandbox/operations.py
```diff
@@ -139,12 +136,27 @@ async def build_single_html(project_id: str) -> str:
...
     dist_dir = os.path.join(workspace_dir, "dist")
-    index_path = os.path.join(dist_dir, "index.html")
-
-    if not os.path.isfile(index_path):  # noqa: ASYNC240
+    if not os.path.isfile(os.path.join(dist_dir, "index.html")):  # noqa: ASYNC240
         raise BuildError(f"Build failed or dist/index.html not found.\n\n{build_output}")
+    return dist_dir
+
+
+@observe(name="build-dist")  # type: ignore[untyped-decorator]
```
88roy88: tbh, I dont think we need this observe

backend/src/flow44/sandbox/operations.py
```diff
@@ -139,12 +136,27 @@ async def build_single_html(project_id: str) -> str:
...
+async def build_dist(project_id: str) -> list[tuple[str, bytes]]:
+    dist_dir = Path(await _build_dist_dir(project_id, vite_base="./"))
+    return [
+        (p.relative_to(dist_dir).as_posix(), p.read_bytes())
+        for p in dist_dir.rglob("*")  # noqa: ASYNC240
+        if p.is_file()
+    ]
+
+
+@observe(name="build-single-html")  # type: ignore[untyped-decorator]
```
88roy88: same

backend/src/flow44/logic/publish.py
```diff
@@ -0,0 +1,34 @@
+import re
```
88roy88, line 1: note sure this is a "service" dont see why its under the services folder. I think this is belongs in the project logic.

backend/src/flow44/services/publish_service.py
```diff
@@ -0,0 +1,34 @@
...
+
+async def publish_project(project_id: str, slug: str | None) -> str:
+    files = await build_dist(project_id)
+    await s3_storage.deploy_dist(project_id, files)
+    handle = slug or project_id
+    await update_project_published_url(project_id, handle)
+    return handle
+
+
+async def delete_published_project(project_id: str) -> None:
```
88roy88: no reason for this function, the `s3_storage.delete_published_prefix(project_id)` worked fine

backend/src/flow44/api/export.py
shalom-raichman: done

backend/src/flow44/sandbox/operations.py
shalom-raichman: done

backend/src/flow44/sandbox/operations.py
shalom-raichman: done

backend/src/flow44/logic/publish.py
shalom-raichman: done

backend/src/flow44/services/publish_service.py
shalom-raichman: done

backend/src/flow44/integrations/s3.py
```diff
@@ -68,29 +79,48 @@ async def _ensure_bucket(self, bucket_name: str) -> None:
...
-            Key=self._key(project_id),
-            Body=html_content.encode("utf-8"),
-            ContentType="text/html",
-            ACL="public-read",
-            StorageClass=settings.S3_STORAGE_CLASS,
+    def _prefix(project_id: str) -> str:
+        return f"published/{project_id}/"
+
+    async def deploy_dist(self, project_id: str, files: Iterable[tuple[str, bytes]]) -> None:
+        await self.delete_published_prefix(project_id)
```
yoni5353: instead of deleting the old project, i think we'd want to mark all those files for deletion (with some S3 Tag) and leave them there for a few days, so that users with an old version of the published app won't get errors of missing images and lazy routes.

i think defining such a lifecycle policy ("if tagged-for-deletion, expire after 7 days") should be easy enough in aioboto, but we can implement it in another branch, imo it's still better to save a bit more files so that users won't get unexpected errors.

backend/tests/logic/test_publish.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.mark.asyncio
+async def test_resolve_slug_status_taken():
+    with patch("flow44.logic.publish.is_handle_taken", AsyncMock(return_value=True)):
+        assert await resolve_slug_status("my-slug", "proj-1") == SlugStatus.taken
+
+
+@pytest.mark.asyncio
+async def test_resolve_slug_status_available():
+    with patch("flow44.logic.publish.is_handle_taken", AsyncMock(return_value=False)):
+        assert await resolve_slug_status("my-slug", "proj-1") == SlugStatus.available
```
yoni5353: you're testing a function that just has if elif else, i think it's a bit redundant

backend/src/flow44/api/shared.py
```diff
@@ -1,46 +1,42 @@
...
 
-def _s3_response(html: str, headers_src: httpx.Headers) -> HTMLResponse:
-    headers: dict[str, str] = {"Cache-Control": f"public, max-age={settings.S3_CACHE_TTL}, must-revalidate"}
-    if "etag" in headers_src:
-        headers["ETag"] = headers_src["etag"]
-    if "last-modified" in headers_src:
-        headers["Last-Modified"] = headers_src["last-modified"]
-    return HTMLResponse(content=html, headers=headers)
+def _cache_headers(asset: S3Object) -> dict[str, str]:
+    headers = {"Cache-Control": f"public, max-age={settings.S3_CACHE_TTL}, must-revalidate"}
```
yoni5353: if im not mistaken this is not good, because index.html will get these cache headers as well, and will be treated as fresh for the TTL even if it's already changed (especially bad if we change assets around it, so it won't be consistent) 

index.html should have no-cache

and stuff inside assets/ might as well have the `Cache-Control: public, max-age=31536000, immutable` because vite hashes them

backend/src/flow44/integrations/s3.py
shalom-raichman: this is a bugfix branch, so I don't want to change the behavior here.
why would users get errors ?

backend/src/flow44/api/shared.py
shalom-raichman: the index have it's own route with no cache, for the asset I don't really get what you are saying

backend/tests/logic/test_publish.py
shalom-raichman: i rewrote the test

backend/src/flow44/integrations/s3.py
shalom-raichman: opened an issue, it's not related for this branch

backend/src/flow44/api/shared.py
shalom-raichman: done

backend/src/flow44/integrations/s3.py
yoni5353: the lifecycle policy (or any other solution) is not related to the branch, so the issue is fine.

but this line is still problematic in this branch, because it can cause breakage that was not possible before. in this branch, remove this line, or make it toggle a flag instead of deleting.

shalom-raichman: Done. Republishing no longer deletes files. It tags the ones the new build dropped as pending-deletion, and a bucket lifecycle rule deletes them after 7 days (configurable).

One catch: S3 counts those days from the last upload, not from tagging, so the tag is added by copying the file onto itself to restart the clock. I also added the lifecycle rule in this branch, so I'll close the issue.

yoni5353 (approved):

PR #190, opened 2026-08-09
chore: fix orphaned test DBs, expose minio ports, apply ruff autofixes

shalom-raichman:
Small unrelated fixes pulled out of a feature branch.

**Orphaned test DBs.** `DatabaseJanitor.init()` does a bare `CREATE DATABASE` while `drop()` uses `IF EXISTS`, so an aborted pytest run leaves `test_flow44_<worker>` behind and every later run fails that worker with `DuplicateDatabase`. Three orphans were breaking 41 tests locally. Now drops before creating, suppressing `InvalidCatalogName` when absent (`drop()` can't be called unguarded — it does `ALTER DATABASE` first).

**Minio ports.** Published 9000/9001 to the host for local dev.

**Ruff autofixes.** Six test files + conftest imports. Mechanical; they were being reapplied every run.

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -76,7 +76,7 @@ async def setup_test_db():
         version="16",
         password=settings.DB_PASSWORD,
     )
-    # An aborted run never reaches __exit__, and init() has no IF NOT EXISTS
+    # Pre-drop the DB if it exists from prior aborted runs (suppresses error if it doesn't exist yet).
     with contextlib.suppress(psycopg.errors.InvalidCatalogName):
         janitor.drop()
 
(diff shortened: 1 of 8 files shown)


backend/tests/conftest.py
```diff
@@ -60,14 +68,19 @@ async def setup_test_db():
...
+    janitor = DatabaseJanitor(
         user=settings.DB_USER,
         host=settings.DB_HOST,
         port=settings.DB_PORT,
         dbname=dbname,
         version="16",
         password=settings.DB_PASSWORD,
-    ):
+    )
+    # An aborted run never reaches __exit__, and init() has no IF NOT EXISTS
```
yoni5353: it took me some time to understand this, because the comment explains the problem and not the solution.
```suggestion
    # Pre-drop the DB if it exists from prior aborted runs (suppresses error if it doesn't exist yet).
```

yoni5353 (approved):

PR #191, opened 2026-08-09
fix: reconcile task contracts in Execute so undelivered files are caught

shalom-raichman:
A task's `files` list is a contract, but nothing checked it. `_execute_task` set `status="completed"` unconditionally after a write loop that may have written nothing, so an undelivered file only surfaced later as an opaque rollup error that never names the failing task. The fix agent then tried to invent it — worst traced run was 10 rounds / 317K tokens.

Across 81 tasks in 13 traced runs, 5 delivered none of their contracted files (4 crashed mid-stream, 1 silently produced nothing). Every project with an unresolved-import build error had at least one such task; no project with all tasks fulfilled had one.

## Changes

- `_unfulfilled` compares `task.files` against `task_files` by set membership — status lies, delivery doesn't
- each gap gets one retry with the contract narrowed to the missing paths and an instruction naming them, which also shortens output enough to reduce the repetition loop behind 4 of the 5 failures
- retries stay at task level; re-entering the step would reset `completed_files` and wipe finished work
- anything still missing is marked `failed` and named to the fix agent — but only alongside a real build error, so a green build is never dragged into the fix loop
- codegen rule: a `.ts` file must not contain JSX. A hook planned as `.ts` returned JSX and caused 6 identical fix rounds; the executor can't rename it (off-contract paths are dropped), so restructuring is its only option
- `fix_error`: rejections alone no longer gate the retry loop

## Verification

`tests/ai/agents` — 39 passed, 7 new covering the retry, `task_files` accumulation, failed-status marking, and the validate note. Ruff and format clean; mypy shows only the known Windows-only `pty.py` errors.

Not yet measured: the clean-build rate against the 4/13 baseline. Draft until that benchmark runs.


shalom-raichman:
not sufficient fix, we'll wait for real data

PR #192, opened 2026-08-12
Shalom/composer restyle

shalom-raichman:
The prompt input before:
image
image

vs

The prompt input after:
image
image


yoni5353 (approved):
everything looks fine, though this will change again anyway when we refactor the add-data-source button like the team discussed

PR #196, opened 2026-08-17
chore: pin litellm version to work on windows

shalom-raichman:


yoni5353:
it's a shame to keep the version locked while we still use it. it just needs regular old windows c build tools, from what i gather. you can download them here as a standalone iirc https://visualstudio.microsoft.com/visual-cpp-build-tools/

PR #200, opened 2026-08-24
Increase run timeout (was 30s by default, increase to 25 minutes)

yoni5353:


backend/src/flow44/integrations/flapi/client.py
```diff
@@ -17,6 +17,9 @@
 
 __all__ = ["FlapiClient", "FlapiUpstreamError", "data_source_client"]
 
+# Flapi's run timeout is 25 minutes
+DEFAULT_FLAPI_CLIENT_TIMEOUT_S = 25 * 60 + 10
```
shalom-raichman, line 21: we already have a FlapiSettings in the config.py and we using it in the next line, lets move it there with the default

yoni5353: fair enough, moved it there

backend/src/flow44/config.py
```diff
@@ -95,6 +95,7 @@ class FlapiSettings(BuildAppSettings):
     # FLAPI base URL. In dev you can point to the local mock (default).
     FLAPI_BASE_URL: str = "http://localhost:6001"
     FLAPI_VERIFY_SSL: bool = True
+    FLAPI_CLIENT_TIMEOUT_SEC: int = 25 * 60 + 10  # Flapi's default run timeout is 25 minutes
```
shalom-raichman, line 99:
```suggestion
    FLAPI_CLIENT_TIMEOUT_SEC: int = 25 * 60 + 10
```

shalom-raichman (approved):

PR #202, opened 2026-08-26
fix: reduce first-build failures from timeouts and JSX in .ts files

shalom-raichman:
57.6% of generated miniapps fail their first build (33 prod traces, Opik `flow44`). Two causes, both fixed here.

**Streaming timeout.** `stream_chat` passed `AI_REQUEST_TIMEOUT` (300s) to a streaming call as a total cap, so generations still producing tokens got killed. Failed tasks leave files unwritten, and dependents then import paths that never exist. Adds `AI_STREAM_TIMEOUT = 600` for the streaming path only; 300s stays for the non-streaming paths. Capped at 600 because `AGENT_RUN_TIMEOUT` (1800) covers all layers of a run.

**JSX in `.ts`.** No prompt stated a `.ts`/`.tsx` rule. Vite resolves `.ts` before `.tsx` and the parser has no rename/delete, so the fixer's corrected `.tsx` stays shadowed — one trace burned all 10 attempts. The extension is chosen by the planner (`architecture` → `merge` → `task.files`), so the rule goes there, plus `codegen` as a backstop.

Narrows the timeout window rather than closing it — past 600s the same failure mode returns. Residual-failure handling, npm EACCES, and vite `resolve.extensions` are out of scope.

**Tests:** `pytest tests/ai` 150 passed; ruff + mypy clean. The new provider test asserts the stream timeout is used and not `AI_REQUEST_TIMEOUT`, so it fails on revert.

### DIFF
--- a/backend/src/flow44/config.py
+++ b/backend/src/flow44/config.py
@@ -62,6 +62,7 @@ class AIModelSettings(BuildAppSettings):
     AI_BASE_URL: str | None = "http://flow-44-models.com/openai/v1"
     AI_API_KEY: str | None = "default"
     AI_REQUEST_TIMEOUT: int = 300
+    AI_STREAM_TIMEOUT: int = 600  # kept well under AGENT_RUN_TIMEOUT, which spans every layer of a run
 
     # if ai_model starts with bedrock/ set base_url and api_key to None
     # (using pydantic v2's model_validator to allow dynamic defaults based on other fields)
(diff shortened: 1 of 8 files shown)


backend/src/flow44/config.py
```diff
@@ -62,6 +62,7 @@ class AIModelSettings(BuildAppSettings):
     AI_BASE_URL: str | None = "http://flow-44-models.com/openai/v1"
     AI_API_KEY: str | None = "default"
     AI_REQUEST_TIMEOUT: int = 300
+    AI_STREAM_TIMEOUT: int = 600  # kept well under AGENT_RUN_TIMEOUT, which spans every layer of a run
```
88roy88: Shalom, do me a favor, no ai comments where they are not needed

88roy88 (approved):

backend/src/flow44/config.py
shalom-raichman: done

